Issue #439, opened 2014-06-12
Expand doesn't take Aliases into account


Yopai:
Let's asume I have a table _declinaison_, each one having one or more _prix_ :
```sql
CREATE TABLE declinaison(
   id INT(8),
   name VARCHAR(100),
   PRIMARY KEY (id)
);

CREATE TABLE prix (
   id INT(8),
   dt_debut DATETIME,
   id_declinaison INT(8),
   PRIMARY KEY(id),
   CONSTRAINT fk FOREIGN KEY (id_declinaison) REFERENCES declinaison(id)
);
```
I wan't to get all the declinaison with their associated prices : the normal one (dt_debut IS NULL) and the promo one (dt_debut IS NOT NULL, and < now)

```php
$nPrixNormal = QQ::Alias(Declinaison::QQN()->PrixAsId, 'prix_normal');
$nPrixPromo = QQ::Alias(Declinaison::QQN()->PrixAsId, 'prix_promo');
$arrProduits = Declinaison::QueryArray(
	QQ::All(),
	QQ::Clause(
		QQ::Expand($nPrixNormal, QQ::IsNull($nPrixNormal->DtDebut, $now)),
		QQ::Expand($nPrixPromo, QQ::LessOrEqual($nPrixPromo->DtDebut, $now))
	)
);
```
=> error : You have two different Join Conditions on the same Expanded Table
Well, that's not the same table, but two aliases.

Moreover, if I only expand on $nPrixPromo, using $nPrixNormal in a where, for example, I got a "invalid column in the ON clause", because the join condition is set on the wrong LEFT JOIN.

olegabr:
merged

PR #441, opened 2014-06-12
2 examples and test case for bug #439 (merged in #440)

Yopai:


olegabr:
re-maked for v3.0 in #877

PR #575, opened 2015-01-17
QDataGridBase documentation. QDateTime legacy class file dropped.

vaibhav-kaushal:
QDataGridBase documentation is now done. Also, QDateTime.Legacy class file has been removed (since we are imposing PHP version 5.4 as the minimum requirement and that file was present as workaround for PHP 5.2 and below)


vaibhav-kaushal:
@spekary Please write the PHPDoc comments for the new classes you have added in the framework's codegen system (QMeta*.class.php files). You know those files the best!

vaibhav-kaushal:
I will be merging the pull requests within the next 2 days in case no one else does that (or points out a mistake).

spekary:
I will be doing that at one point. I am going to be changing the name as part of the change of the MetaControl name to something like ModelAdapter.

PR #577, opened 2015-01-21
feature-QControlBaseRefactor

spekary:
Refactoring QControlBase attributes and related "Style" classes. Way too much duplicate code there, and too much unnecessary stuff in the form state. This also will make it easier for CSS frameworks like Bootstrap, since they heavily manipulate attributes through classes and date-* attributes. I am hoping this can be a really solid base class for future improvements.

To summarize the results of the refactoring:
1. A new class underneath the QControlBase that manages attributes and styles in arrays. This will be much more flexible for any future HTML changes that we can't predict. This takes over the management of all the repeated code in QContorlBase, QListItemStyle and QDataGridRowStyle.
2. A couple of utility routines to have the HTML print out so its more readable and easier to debug, especially in the javascript. A __MINIMIZE__ define can easily remove these extra spaces if desired. The QHtml class replaces the QCss class, and includes additional utility code to draw html tags.
3. Refactoring how ajax commands are sent into qcubed.js. They are now sent as json, so future changes to this will be easier to make, and the code easier to understand.
4. Removal of some outdated or obsolete code, including:
- attributes that are not supported in HTML5.
- drawing artifacts left over from ancient debug code
- function parameters that are never used
5. An effort to move towards PSR-1 method names. Going to lower-case camelCase will not impact legacy code, since PHP calls methods in a case insensitive way.

This effort is incomplete. This is an intermediate point in the process, and implements the changes made in QControlBase into the QTextBox, QListControl and QControlProxy classes as an example of the impact. The other controls need to be updated.

@olegabr and @vaibhav-kaushal, would appreciate your comments on this before I continue, and whether you think I am on the right track.


vaibhav-kaushal:
I think it is brilliant except one thing: 

> An effort to move towards PSR-1 method names. Going to lower-case camelCase will not impact legacy code, since PHP calls methods in a case insensitive way.

The problem is again - consistency. There are two parts: 

1. We are almost *used to* typing the way it is. Introducing new javaCasing against CamelCasing will break it. Not that I am *against* this but it really takes a shift. It breaks class/property/method naming conventions too (right now all of them are CamelCased).

2. If we are trying to make that shift, it better be in one go, which means it's best we do it in only a few commits and only after we are done with 3.0 release (so that would push this one to at least 3.1). Introducing  feature which needs a change as big as this (renaming all methods across framework is not an easy task) in partial manner among other changes is not a wise idea, I suppose. 

And why exactly are we trying to make the method names PSR complaint (as in - what is wrong if we do not adhere to those rules?)
----

Apart from that, other changes are great! I have long thought about improving QCss myself. Replacing it and creating a full fledged class which can handle it's own stuff is great. However, I have a suggestion - the previous hierarchy should be restored and the new QHtmlAttributeManager class should be derived from QBaseClass instead. Reasons: 

1. Doing so will make sure that we can use the attribute manager to create an individual set of html attributes. 
2. We can create a separate variable on these 3 classes which can hold null or instances of HtmlAttributeClass. This will reduce the FormState size substantially.
3. We can still leave all the properties in place. All we have to do is something like this:

```
case "FontSize":		
		try {		
			$this->objHtmlAttributes->FontSize = QType::Cast($mixValue, QType::String);
			break;		
		} catch (QInvalidCastException $objExc) {		
			$objExc->IncrementOffset();		
			throw $objExc;		
		}
```

The rest seems good.

spekary:
I understand the arguments on the camelCase. There is already some code that is lower camelCase in the core, so I think its consistently inconsistent :-). In other words, should we go back and upper case all the other inconsistencies, if we know we are headed towards the lower case version eventually?

The reason to go PSR-1, 2 is to continue down the path of trying to get more people to use the framework and remove barriers to them using it. @scottux asked for this a while ago, and started a project to do it. PSR-1 and PSR-2 are not huge changes. namespacing with PSR-4 would be a little bigger change.

Not quite sure what you meant on the second part. QHtmlAttributeManager derives from its base class, which derives from QBaseClass, so that is already there. See the QTagStyler for a derived class that will hold a collection of attributes that can be passed around.

Your suggestion 3 recreates the problem this was trying to solve. Creating a big collection of __set and __get items that are copies of what are already in the QHtmlAttributeManager and that simply hands off the function creates extra maintenance work. I am trying to reduce code duplication. 

Ideologically, the hierarchy is saying that something like a QTextBox "is a" QHtmlAttributeManager, rather than saying that it "has a" QHtmlAttributeManager. By suggesting that the attribute names should be put in both places, you are implying that it passes the "duck" test, as in, it walks like a duck, talks like a duck, so its a duck.

vaibhav-kaushal:
Yeah, inconsistently consistent (or the other way around) ;). But yes, I would suggest that before we attempt on going full PSR-1 compliance, it is better to get them all uppercase. As far as more users are concerned, PSR compliance on naming conventions are not that great an attraction. They can be attractive only for someone who values usability less than standards. This is a call we should take later (I’m not against it, but right now, mixing in more javaCase functions would be a little confusing). However, that should go into another commit for sure.

About QHtmlAttributeManager derives, I would like to elaborate about my vision about it:

Right now, we say something like this to control:

```
$this->lblMessage->FontSize = '20px';
```

We should change it to something like: 
```
$this->lblMessage->objStyle->FontSize = '20px';
```

For that to happen, we will have to create a ```objStyle``` for ```lblMessage```. Now that is not possible unless *we have a* ```objStyle```. I know that this is something you do not like because of the apparent performance drops of ```__set``` and ```__get```. However, you would be calling those methods anyway (yes, down on the QHtmlAttributeManager level, but it's there anyway). For the QTagStyler part - We can simply merge QHtmlAttributeManager into it (or vice versa). 

I know I am repeating myself. But the point I am trying to make is - if we keep QHtmlAttributeManager separate, I can generate HTML attributes separately, save them in a QForm-wide variable and apply it to more than one label at runtime or create a set of HTML attributes to be used in the template file. It will reduce the QFormState size for even very complex pages (because you are doing away with a chunk of attributes which are always present in the state of that variable).

And the point 3 in my previous message - it is an optional one, only for backward compatibility. However, that can be dropped if that is in interest of the framework.

Creating just another layer does not appeal to me when an alternate solution solves the problem and adds more features too! I hope I am making sense.

spekary:
I think you can achieve your goal in a different way. Doing what you are describing will cause a lot of current code to break. It also violates the idea of OOP. A QControl doesn't have an attribute manager, it IS an attribute manager because it HAS attributes. The QHtmlAttributeManager is a foundational class that isn't meant to be publicly consumed, similar to QBaseClass. Note that it manages all kinds of attributes, not just css styles. For example, you wouldn't want to do 
```
$objTextBox-->objStyle()->AriaLabelBy = $strLabelId;
```
AriaLabeledBy should be something you can directly set on a control without being forced to go through a styler, especially since its not logically a style thing.

Now, there is a QTagStyler object as part of this pull request, which does use the QHtmlAttributeManager and is there to do just what you are talking about. You can use it to store a repository of various style things and attribute things, and then tell any QControl to use it as an override.

olegabr:
Great attempt and very interesting discussion!
But let's "dance from the stove" (russian proverb). It means that we should make our reasoning from the framework goals. @spekary talks about some PSR-s, @vaibhav-kaushal talks about some optimisations. All of that is good things, but none of that are the qcubed framework goals ;-)

The main goal is stated there: https://github.com/qcubed/framework/wiki/Roadmap#preamble

"The mission of QCubed is to be a rapid web application development framework targeted to web development beginners. The technologies a user must to know in order to develop with QCubed are very limited:
 * Basic SQL
 * PHP
 * Basic HTML

This beginner-friendliness of QCubed is its main power and must be maintained carefully. All additions to, and modifications of, the QCubed core must be weighted with this friendliness in mind."

What does that mean for that particular discussion?

First, we should not sacrifice this simple syntax:
```php
  $this->lblDescription->BorderWidth = "1px";
```
It is important for beginners, and it is important for fast prototyping.

Second, if we can improve performance in some way, we should do that.

Third, if we can provide a way for the beginner to learn the WEB with qcubed, we should do that. By saying that I have the following project and beginner programmer evolution flow in mind:
 - fast prototyping with code like the above;
 - refactor it to the `page.tpl.php` file as a CSS rules;
 - move the CSS code from the `tpl` file to some project-specific CSS file in the `assets/css` folder.

The second and the third steps require from the WEB beginner some CSS learning and motivation. It was exactly the way that me and my qcubed-based project evolves in time.
But can the qcubed help the beginner to complete these steps faster? Lets see.

What if qcubed, instead of storing all of these CSS attributes, whould just generate a CSS file for the page? Let it be `assets/css/generated/page.css`. The file would contain some generated classes used by controls on the `page.php`. The generation process is as follows: combine all CSS attributes set in code for some control, minify it and calculate some hash value from it, e.g. md5. Use it as a CSS class name. Render the control with this class name in it's class html attribute. In a FormRender method combine all of these CSS code into one string, calculate a hash from it and compare with the hash previously calculated (and cached!) for the `assets/css/generated/page.css` file already stored on the filesystem.

With this process, qcubed developer would always have a working CSS file for the `page.php` developed. OK, the prototype is ready and he wants to improve it for a production. What he can do?
 - copy the contents from the `assets/css/generated/page.css` to the `assets/css/page.css` file
 - rename class names to be meaningfull (provided control ids in css comments would be helpfull here)
 - add these CSS classes to controls in the page.php
 - switch the qcubed to the production mode. e.g. switch the SERVER_INSTANCE constant from `dev` to `prod`

As a result:
 - the `page.php` is clear of css attributes management;
 - the beginner programmer have a working CSS ready to experiment with
 - the `page.php` automatically loads the file `assets/css/page.css`

@spekary, you have called for the right track. It is the right one, I believe ;-)

vaibhav-kaushal:
@olegabr has got some neat ideas! Its nice. But I am going to build the styles manager myself and see if it works out (can put it in QControl so it does not conflict with the rest of the framework). Only after that can this debate have more meaning.

spekary:
@olegabr I agree with that line of reasoning. The only caveat is "specificity" in CSS. Styles put it a css file are not as specific as ones attached to an object. But that is a minor point. The idea of being able to do quick modeling in code, and then have it generate CSS is great, and should be easy since it already will generate the css.

@vaibhav-kaushal, please take a quick look at the "override" method in QHtmlAttributeManagerBase. Most of what you want is already there. I tried a number of ways of doing this, with a trait, and with a separate object as you suggested, but in the end, having an additional base class made the most sense to me and was most flexible. The other methods required a lot of duplicated code and weird handoffs. 

We cannot predict how HTML will develop. That is why I created a QHtmlAttributeManager for the project/controls directory, and a QHtmlAttributeManagerBase in the core. That way, the developer can add additional base attribute functionality without worrying about changing a core file, and then that functionality becomes a candidate for inclusion in the core later. Also, it makes it possible for plugin css frameworks to inject additional functionality by having the project classes inherit from the plugin that then inherit from the core. This "dependency injection" is getting some attention in various frameworks. And while this method of doing dependency injection isn't automatic, in that developers will need to modify their local "controls" files rather than having some kind of mechanism for plugins to insert themselves, it is easier to understand, easier to support, and faster since it will take advantage of the PHP pre-processor, and ultimately will take advantage of the HHVM pre-processing as well.

How things are written now, you  can do the following:

```
$objStyler = new QTagStyler();
$objStyler->FontSize = 12;
$objStyler->ForeColor = 'gray';
$objStyler->CssClass = 'myclass';

$objTextBox = new QTextBox($this);
$objTextBox->override ($objStyler); // give all the objStyler attributes to the control
```

What this will not do is allow for making changes to $objStyler after the override, and then having those changes automatically affect the $objTextBox. I know QCubed allows that kind of things in some places, like QDataGridRowStyler, but I think that is a really bad idea as a general practice for all objects. A basic programming practice is to not have hidden dependencies, such that changing one object will invisibly impact other objects. Doing things this way is taking hidden advantage of how PHP implements objects with references. It works OK for QDataGridRowStyler and QListItemStyler because those conceptually are repeated objects within the base object, they are not the base object themselves.

spekary:
This should be ready to merge. I made sure all the controls that needed to were using the new HTML generator. The only one not using it is QDataGrid, but it does its own thing so that is OK. I also put everything back to upper CamelCase that I could find, per @vaibhav-kaushal suggestion.

vaibhav-kaushal:
I have no time to review these days. Extremely busy with something. Sorry for not being available. If it is ready, its good to merge.

PR #580, opened 2015-01-21
More documentation (mostly for QDatabaseBase)

vaibhav-kaushal:
More documentation (mostly for QDatabaseBase). Will probably halt for next few days.


vaibhav-kaushal:
Just wanted to ask: when can these API Docs that I am adding will become live on the official QCubed site?

spekary:
Probably when we get 3.0 released.

PR #587, opened 2015-02-02
bugfix-CodegenErrors

spekary:
Fixes a problem where a template that calls a non-existant method would cause codegen to return a blank page, rather than an error code.

Also restored title on load of metacontrol.

### DIFF
--- a/includes/framework/QExceptions.class.php
+++ b/includes/framework/QExceptions.class.php
@@ -67,7 +67,9 @@
 		 * at the very least the Caller of the most immediate function, which is
 		 * 1 up the call stack, is responsible.  So therefore, by default, intOffset
 		 * is set to 1.
+		 * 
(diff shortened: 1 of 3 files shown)


includes/framework/QExceptions.class.php
```diff
@@ -171,6 +173,15 @@ public function __construct($strType, $strClass, $strProperty) {
 	}
 
 	/**
+	 * Thrown when we try to call an undefined method. Helpful for codegen.
+	 */
+	class QUndefinedMethodException extends QCallerException {
+		public function __construct($strClass, $strMethod) {
+			parent::__construct(sprintf("Undefined method in '%s' class: %s", $strClass, $strMethod), 2);
```
olegabr: can localization be applied here?

spekary: Yes, good catch. It should be done.

PR #588, opened 2015-02-03
feature-ScopeAttribute

spekary:
Adding scope attribute to table headers


olegabr:
just wander of what the "scope" is used for?

spekary:
Its used for assistive readers for the blind and partially blind. JQuery UI will automatically insert various aria attributes and the like into their widgets. Bootstrap shows many examples of adding these manually. Its becoming common practice to include these various special attributes. Its does nothing visually on the screen, and would only impact special browsers designed for the blind.

olegabr:
@spekary thank you!

PR #590, opened 2015-02-03
bugfix-ValidationErrorRedraw

spekary:
Changing the setting of the validation error to the attribute. Fixing a problem where setting the validation error from within controls by just setting the text did not cause a redraw.Also, make a distinction between Errors and Warnings for CSS frameworks.

### DIFF
--- a/includes/base_controls/QDateTimeTextBox.class.php
+++ b/includes/base_controls/QDateTimeTextBox.class.php
@@ -86,7 +86,7 @@
 					$dttTest = QDateTimeTextBox::ParseForDateTimeValue($this->strText);
 
 					if (!$dttTest) {
-						$this->strValidationError = $this->strLabelForInvalid;
+						$this->ValidationError = $this->strLabelForInvalid;
 						return false;
 					}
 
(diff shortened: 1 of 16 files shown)


includes/base_controls/QDateTimeTextBox.class.php
```diff
@@ -100,7 +100,7 @@ public function Validate() {
 						}
 
 						if ($dttTest->IsEarlierThan($dttToCompare)) {
-							$this->strValidationError = 'Date cannot be ' . $strError;
+							$this->ValidationError = 'Date cannot be ' . $strError;
```
olegabr: translation support is required here

includes/base_controls/QDateTimeTextBox.class.php
```diff
@@ -115,7 +115,7 @@ public function Validate() {
 						}
 
 						if ($dttTest->IsLaterThan($dttToCompare)) {
-							$this->strValidationError = 'Date cannot be ' . $strError;
+							$this->ValidationError = 'Date cannot be ' . $strError;
```
olegabr: and there

PR #593, opened 2015-02-04
feature-Refactor of list items

spekary:
Per previous suggestion of @olegabr, this is a refactoring of list items to make QHtmlList a QListControl. Its a little messy at the QListItem level, because it has to serve dual purposes, but it is likely that many javascript widgets that would use an HtmlList as the base would need the QItemList functionality, like having Selected items, Values, and such. The other odd thing for a QListBox is the QListItem can have sub items. However, the HTML spec says that this is a possible future addition to the spec through the optgroup tag, so its not that big a stretch.

QHtmlList is now a QListControl.
QListItem is now a hierarchical item.
QListControls do not expose their value to the client. Since that is often a database id, this was a security risk. Value is still accessible in PHP though.

### DIFF
--- a/assets/php/tests/ui_basic.php
+++ b/assets/php/tests/ui_basic.php
@@ -99,6 +99,7 @@
 			$this->lstSelect->Warning = 'Value = ' . $this->lstSelect->SelectedValue;
 			$this->lstSelect2->Warning = 'Values = ' . implode (',', $this->lstSelect2->SelectedValues);
 			$this->lstCheck->Warning = 'Values = ' . implode (',', $this->lstCheck->SelectedValues);
+			$this->lstCheck2->Warning = 'Values = ' . implode (',', $this->lstCheck2->SelectedValues);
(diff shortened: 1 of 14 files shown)


spekary:
Not sure about this yet. Closing for now.

vaibhav-kaushal:
You are more experienced than me, but I tend to be in favor of 'not messy'. If it's getting messy, then it should be simplified. Also, why does @olegabr want QHtmlList to be a QListControl. Can you please reference to the issue where this was discussed. 

Sincerely

spekary:
#426. QTreeNav is based on QListControl, and I am not sure it should be. It is tricky, because if you just think of these as HTML generators, then they are really different objects. But if you think of them more abstractly, as collections, then their interfaces and the idea of what they are gets closer. But after trying it, I am not sure they are close enough. I am going to do some more experimenting. It may be that an Iterator interface is the right thing here. Not sure that helps or not.

PR #597, opened 2015-02-07
bugfix-CheckboxDisplay

spekary:
Fixing RenderLabelAttributes to correct display problem.

### DIFF
--- a/includes/base_controls/QCheckBox.class.php
+++ b/includes/base_controls/QCheckBox.class.php
@@ -98,7 +98,7 @@
 				if (!$this->blnWrapLabel) {
 					$strLabelAttributes = ' for="' . $this->strControlId .'"';
 				} else {
-					$strLabelAttributes = $this->renderLabelAttributes();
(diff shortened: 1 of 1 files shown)


includes/base_controls/QCheckBox.class.php
```diff
@@ -146,7 +146,10 @@ protected function renderLabelAttributes() {
 			if (!$this->Enabled) {
 				$objStyler->AddCssClass('disabled');	// add the disabled class to the label for styling
 			}
-			return $objStyler->renderHtmlAttributes();
+			if (!$this->Display) {
```
olegabr: may be just
```php
$objStyler->Display = $this->Display
```

spekary: Not in this case. If the user puts a display parameter on the $objStyler to turn it off or do something else, then it will override that and turn it to the default on. I only want to override the display if it is specifically turned off.

PR #598, opened 2015-02-07
feature-ReverseReferenceExpansion

spekary:
Removes the underline in front of the reverse reference expansion so that these expansions can be referred to in data grid columns and the like. Also fixes counting problem in the data binder for these.


includes/codegen/templates/db_orm/class_gen/property_get.tpl.php
```diff
@@ -84,7 +84,8 @@ public function __get($strName) {
 	$varPrefix = (is_a($objAssociatedTable, 'QTypeTable') ? '_int' : '_obj');
 	$varType = (is_a($objAssociatedTable, 'QTypeTable') ? 'integer' : $objReference->VariableType);
 ?>
-				case '_<?= $objReference->ObjectDescription ?>':
+				case '<?= $objReference->ObjectDescription ?>':
+				case '_<?= $objReference->ObjectDescription ?>': // for backwards compatibility
```
olegabr: here and for other changes like this the properties doc template should be changed accordingly

PR #616, opened 2015-02-21
add a configuration option to load the javascript files with <script> tags

vakopian:
This change adds a new configuration option called `__PRELOAD_JS_FILES__`, which allows loading the javascript files with `<script>` tags rather than with `qc.loadJavaScriptFile`.
The current approach of loading control-specific (including plugin) javascript files with `qc.loadJavaScriptFile` creates lots of problems when debugging with firebug or chrome.
By default this flag is off, so the change is fully backward compatible.
The big block of code that shows as a diff, is not new, it's just moved up within the same function.

### DIFF
--- a/includes/configuration/configuration.inc.php.sample
+++ b/includes/configuration/configuration.inc.php.sample
@@ -143,6 +143,10 @@ if (!defined('SERVER_INSTANCE')) {
 			define ('__APP_IMAGE_ASSETS__', __SUBDIRECTORY__ . '/assets/images');
 			define ('__APP_PHP_ASSETS__', __SUBDIRECTORY__ . '/assets/php');
 
+			// When __PRELOAD_JS_FILES__ is set to true, the core and plugin javascript files will be included into the
+			// page with <script> tags, rather than loaded dynamically with qc.loadJavaScriptFile
+			define ('__PRELOAD_JS_FILES__', false);
+
 			// There are two ways to add jQuery JS files to QCubed. Either by absolute paths (Google CDN of
 			// the jQuery library is awesome! It's the default option below) - or by using the jQuery
 			// installation that's local to QCubed (in that case, paths must be relative to __JS_ASSETS__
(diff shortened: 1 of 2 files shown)


vaibhav-kaushal:
I think this is a nice addition. However, I think we can (also?) have a method in QApplicationBase which can be used to produce the lines in head. We could call it this way in the Template file: 

```
echo QApplication::DumpPreloadJsTags();
```

and be done with it.

Would that be bad?

spekary:
Javascript files should not be included in the head. It slows down the loading process. Alpha3 currently only uses loadJavascriptFile for ajax injected controls that are not previously loaded. The rest of the javascript files are included in <script> tags. css files have an option to include them in the head.

vaibhav-kaushal:
Yeah, I forgot about that. And that some scripts might need to use jQuery which is available only at the end of the page. 

Please ignore my previous comment. :(

vakopian:
@spekary, I'm not sure if your comment is an objection to this feature, or just a reply on @vaibhav-kaushal comment. 
Either way, this patch does not put the scripts in the header, it puts them right before the big script block, together with jQuery.js and qcubed.js

spekary:
@vakopian Just a reply to putting the js in the head tags. And, a mention that I believe what you are doing here is already done in alpha3. Whatever we can do to make debugging easier is fine by me :-). 

We should be able to go to beta in version 3 very soon. I have one more major change to make (renaming metacontrol to ModelController). Would love to get you using it so your coding skills could go towards it. Do you think you could document the transition, if I walk you through it?

spekary:
I am committing this, but for alpha3, this should be the default behavior. No option is really needed here. The only caveat is that there is still a need for dynamic loading of javascript files to support those rare situations when a plugin control gets injected by ajax after the initial page load. I foresee a time coming up when we are going to applications doing lots of these injections, and while its not the most performant way to do things, for prototyping purposes and ease of use, its very handy.

vakopian:
FYI - I had to make another commit to fix a bug introduced by this change: 98e592786
@spekary If you had merged this PR into alpha 3, please make sure to also get this commit

PR #617, opened 2015-02-21
feature-NewPersistance

spekary:
Defines a new persistence model. Saves and restores only what is needed to restore the visual state of the control, so very space efficient. Also allows saving and restoring of the entire visual state of the application.


vaibhav-kaushal:
This is a lovely implementation!

*off topic*: I will resume my efforts for documentation once we have moved to beta stage (where I do not expect major changes).

PR #629, opened 2015-02-24
Alpha 3.0 master sync

vakopian:


spekary:
This wasn't ready to merge. Some of these changes were implemented in other ways already in the new alpha3. In particular, the preload js files was not necessary, they are already preloaded. I will try to unravel this.

spekary:
Looks like this didn't change what I thought. OK to merge, though something has been introducing files in the old structure.

Issue #630, opened 2015-02-24
alpha-3: codegen uses in core controls


vakopian:
In alpha-3 many `Codegen_` static functions have been added to core control classes.
I'm not sure why this was necessary, but I don't think QCubed core should have any dependency on codegen. In fact one of my goals is to separate codegen into it's own project separate from qcubed core (and pulled with composer).
Can we move these new methods into some helper classes inside codegen?

vakopian:
I came up with a way to refactor this code to decouple codegen from the core. To see a preview of what I'm proposing please take a look at this https://github.com/vakopian/jqcubed/compare/qcubed:alpha-3.0...model-connector-codegen-refactor
Here are the advantages of this approach:
* core does not depend on codegen
* plugins can still provide custom connectors by creating their own class that extend `ModelConnectorControlHelper` (using the class name convention `ModelConnectorControlHelper_${PLUGIN_CONTROL_CLASS}`)
* The abstract base class `ModelConnectorControlHelper` provides a clear interface of what the custom control needs to implement
* Completely avoid using the `ReflectionClass` and its `hasMethod` facilities.
* Less code in templates when using these helper objects
* Overriding is done with regular class inheritance (rather than the php's static inheritance)

I've only refactored the code for `QControl`, `QLabel` and `QListBox` so far.
If you guys agree with my approach, I'll finish the whole thing and make a pull request.

olegabr:
I like this aproach. Please, show an example of it for some plugin control.

vakopian:
As the _Make Your Own Plugin, Part 1: Writing Custom Controls_ example explains (in `components.php`), to use a plugin control (that does not extend one of the core controls) as part of the code generator, the plugin control class has to define some static methods like `Codegen_ConnectorCreate()`, `Codegen_ConnectorRefresh()`, etc.
With my proposed change, instead of putting these static methods into the plugin control class, the plugin package would have to define a separate class named `ModelConnectorControlHelper_${PLUGIN_CONTROL_CLASS}` that implements the `ModelConnectorControlHelper` interface (of course it could also extend and override one of the existing subclasses in core codegen).

None of the current plugins for alpha3 uses this feature, so I can't really show a good example for an existing plugins. But I definitely like this feature and would use it when I port my projects (which are based on my jQCubed scaffolding) since I use many plugin controls in my custom templates.

vakopian:
BTW, `ModelConnectorControlHelper` is not the best name, so if someone can suggest a better name...

[...]

PR #633, opened 2015-02-25
decouple controls from codegen

vakopian:
As per #630


spekary:
I think this is creating way too many files. Can you put the GetModelConnectorParams back into the main class for each control? That really doesn't need to be decoupled.

vaibhav-kaushal:
Are we adding this into 3.0? Or can it be moved to 3.1?

vakopian:
Please see my last comment in #630, and review #655

PR #639, opened 2015-03-05
QCheckBoxList - added CheckAll(), UncheckAll()

lackovic10:
Add functions `CheckAll()` and `UncheckAll()` to the `QCheckBoxList` class.

### DIFF
--- a/includes/qcubed/_core/base_controls/QCheckBoxList.class.php
+++ b/includes/qcubed/_core/base_controls/QCheckBoxList.class.php
@@ -32,18 +32,18 @@
 	 * @property boolean $HtmlEntities
 	 */
 	class QCheckBoxList extends QListControl {
-		
+
 		const ButtonModeNone = 0;
 		const ButtonModeJq = 1;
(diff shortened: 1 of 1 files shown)


lackovic10:
yes. do you want just one function and passing in the `$selected` parameter ?

spekary:
There is already an UnselectAllItems, so leave that the same. Instead of passing in $selected, pass in blnMarkAsModified to mirror the UnselectAllItems call. In other words, have both functions.

lackovic10:
sorry i don't find the function [UnselectAllItems](https://github.com/qcubed/framework/blob/master/includes/qcubed/_core/base_controls/QListControl.class.php) can you please point me to it ?

spekary:
Its in the Alpha3 branch.

spekary:
I see what happened. I will merge this to 2.2. We will need to do it a little differently for alpha3.

lackovic10:
I will create a new PR for alpha3 like you asked

PR #646, opened 2015-03-08
add QQCondition classes for SQL Exists and Not Exists

vakopian:


spekary:
Usually for this kind of core addition, we ask for an example and possibly unit test to make sure it works, but I will merge anyway. If you can provide either of those, it would be helpful.

PR #655, opened 2015-03-11
Model connector codegen refactor phase1

vakopian:


includes/codegen/controls/AbstractControl_CodeGenerator.class.php
```diff
@@ -0,0 +1,79 @@
+<?php
```
spekary: Why is this called AbstractControl_CodeGenerator and not ControlBase_CodeGenerator to correspond with the ControlBase file?

vakopian: Now that I renamed the classes to have the "Base" suffix, we already have `ControlBase_CodeGenerator` which contains the actual generic implementation. We can make that class the root of this hierarchy, but I kind a like having the (almost) pure abstract class which kinda plays the role of an interface and documents which methods are expected to be implemented.
If you'd like, I could actually make that into a real interface (and move the little code that's there into `ControlBase_CodeGenerator`).

includes/codegen/QCodeGenBase.class.php
```diff
@@ -815,54 +815,62 @@ public function ModelConnectorLabelVariableName($objColumn) {
 		 *
 		 * @param QColumn|QReverseReference|QManyToManyReference $objColumn
 		 *
-		 * @return string Class name of control which can handle this column's data
+		 * @return AbstractControl_CodeGenerator helper object
 		 * @throws Exception
 		 */
-		public function ModelConnectorControlClass($objColumn) {
-
+		public function GetControlCodeGenerator($objColumn) {
```
spekary: You removed ModelConnectorControlClass and replaced it with this. This is more difficult to override in QCodeGen.php and specify a default class for each table type. I would prefer you to reinstate ModelConnectorControlClass, and then base this function on the results of that function. That would make it much easier for a developer to specify default classes for data types.

spekary:
I made a comment on your removing ModelConnectorControlClass. That function was useful to override and specify default control classes for particular data types. The new function is more difficult to use. I would prefer it if you reinstated ModelConnectorControlClass, and then based GetControlCodeGenerator on the results of that function.

Also, QDateTimePicker is still using ModelConnectorControlClass, which no longer exists in this pull request.

Think through how plugins would create a CodeGenerator class as well.

spekary:
An additional issue is that all the non-base generators should be put in install/includes/codegen/controls so that they are not over-written when updating the core using Composer.

vakopian:
>  That function was useful to override and specify default control classes for particular data types. The new function is more difficult to use.

In what way is it more difficult to use? You can still override the new method `GetControlCodeGenerator` and return something else for a particular data type.

> I would prefer it if you reinstated ModelConnectorControlClass, and then based GetControlCodeGenerator on the results of that function.

I actually implemented both approaches to compare (even before I did this pull request). The reason I wanna stick with the new approach is because the new approach uses the actual types to instantiate the objects. With the old approach, even for the built-in controls like `QTextBox` we'd have to do:
```
$strClassName = $this->ModelConnectorControlClass($objColumn) . '_CodeGenerator';
return new $strClassName();
```
So if you search for usages of `QTextBox_CodeGenerator` in PhpStorm (or any other tool), it will not find anything. So unless you know where to look, there is no way to discover how these classes are used.
On top of the usage search, you also get better navigation in PhpStorm, because with `new QTextBox_CodeGenerator()` you can just ctrl-click and go to the `QTextBox_CodeGenerator`. Even without PhpStorm, it's kinda the whole "strongly typed vs stringly typed" discussion.
I understand we still do it for custom controls and for plugins, but at least the core control code generator classes are "discover-able".

If you absolutely want the old `ModelConnectorControlClass`, method, then we could compromise on code like this
```
$strClassName = $this->ModelConnectorControlClass($objColumn);
switch ($strClassName) {
    case 'QLabel':
        return new QLabel_CodeGenerator();
    case 'QTextBox':
        return new QTextBox_CodeGenerator();
    // ...
}
```
That way we keep your original method, but we also have my IDE-friendly/strongly-typed way. What do you think?

> Think through how plugins would create a CodeGenerator class as well.

I thought we've already talked/clarified this in #630 (see my last comment). You just need to create a `{PLUGIN_CONTROL}_CodeGenerator` class and register with the class autoloader. 

> Also, QDateTimePicker is still using ModelConnectorControlClass, which no longer exists in this pull request.

Not sure what you mean, I must be missing something. The pull request does have the new `QDateTimePicker_CodeGenerator` class, which is instantiated correctly. I also did a global search for `ModelConnectorControlClass` and could not find anything in my pull request branch. Could you add a line comment in the pull request, so I can find what you're referring to?

vakopian:
> An additional issue is that all the non-base generators should be put in install/includes/codegen/controls so that they are not over-written when updating the core using Composer.

Done with the new commit 4f5e360

spekary:
There are a couple of things lost in the current implementation:
1. An easy way to associate a database type with a  QControlType. The previous function made it easy for a novice to QCubed to specify the mapping, without needing to know about CodeGenerators.
2. More importantly, you are requiring that the database types be associated with a real code generator. So, now I can't associate just any control with a db type, it has to have a generator,  unless I go through the ModelConnector editor.

For example, in my current QCodeGen.class.php file, I have this:
```
				case QType::DateTime:
					if ($objColumn->DbType == 'Date') {
						return 'QDatepickerBox';
					}
					elseif ($objColumn->DbType == 'Time') {
						return 'QTimepickerBox';
					}
					else {
						return 'QCubed\Plugin\QDateTimePickerBox';
					}
```

Since you have this code executing before that:
```
					while (!class_exists($strControlCodeGeneratorClass)) {
						$strControlClass = get_parent_class($strControlClass);
						if ($strControlClass === 'QControl') {
							throw new QCallerException("Cannot find an appropriate subclass of AbstractControl_CodeGenerator for ".$strOrigControlClass);
						}
						$strControlCodeGeneratorClass = $strControlClass .'_CodeGenerator';
					}
					return new $strControlCodeGeneratorClass($strOrigControlClass);
```
I will get an error, because it will not look up the correct code generator. I have to know ahead of time what code generator to use. And if something changes in the code generator chain that introduces additional code generator classes, it might not pick that up. This would introduce a subtle, difficult to debug and difficult to understand bug for someone not expert in QCubed inner workings.

> So if you search for usages of QTextBox_CodeGenerator in PhpStorm (or any other tool), it will not find anything.

I don't see how the two are related. You are already building the `$strControlClass .'_CodeGenerator'` string. I can search for `QTextBox_CodeGenerator` just fine, whether this code is there or not. I understand that since the code generators would not be explicitly called out in this function, you can't just click and go to them, but that minor issue should not trump losing major functionality like the 2 issues I mention above. I also understand about the desire to not have strings contain the class name, but for codegen, I think its OK. 

Maybe another way around his would be to have a static method in the QControlBase itself called "GetControlGenerator", that then returns a code generator for that control, and can be overridden. Actually, I like that way of doing it a lot. You then don't need to do all that work to hunt through the parent chain. You just get the name of the control, and then call:

```
$strControlName::GetControlGenerator();
```

And let inheritance do the rest. So, QTextBox's GetControlGenerator() would be this:
```
static function GetControlGenerator($strOriginalClass) {
    return new QTextBox_ControlGenerator($strOriginalClass);
}
```

Other than this issue, the rest of it seems to be working well. So, we are close.

vakopian:
About your first point, I don't understand what's the difference between knowing that you need to return the name of the type or knowning that you need to return the corresponding _CodeGenerator object. If the developer is using his own type, he will need to write the code for the code generator anyway. In fact it's much cleaner to write it in a separate class with a well defined interface. Once that code is written, it's only natural that the class should be instantiated and returned. If the developer is trying to use an existing type, again it's much easier to return a new object of the _CodeGeenrator type (in fact the editor will even help you with autocomplete).

About your second point, I still don't understand where is the difficulty with overwriting. I'll modify your code in QCodeGen class for the new way:
```
public function GetControlCodeGenerator($objColumn) {
    switch ($objColumn->VariableType) {
        case QType::DateTime:
              if ($objColumn->DbType == 'Date') {
                  return new QDatepickerBox_CodeGenerator();
              }
              elseif ($objColumn->DbType == 'Time') {
                  return new QTimepickerBox_CodeGenerator();
              }
              else {
                  return new \QCubed\Plugin\QDateTimePickerBox_CodeGenerator();
              }
              break;
    }
    return parent::GetControlCodeGenerator($objColumn);
```
What's the problem with this? Your special cases should never get into the while loop in the parent class at all, because you handle them before falling back to the parent's method.

And finally about letting it to the `$strControlName::GetControlGenerator()` method to determine the code generator. This wouldn't achieve the main goal, which is keeping the core independent from the codegen.

spekary:
> This wouldn't achieve the main goal, which is keeping the core independent from the codegen.

Fundamentally I disagree with this goal. Trying to achieve this goal creates more work and achieves less, and is forcing you to have to walk the inheritance tree instead of just letting PHP do it through normal inheritance.  The QControls should know how to codegen themselves. Also, the `GetControlGenerator()` function would now be the only hook into the codegen process and would only have one line of code, so it seems like a very small thing.

Regarding your rewrite, there is no `QTimepickerBox_CodeGenerator` class. You would be forcing me to create one that would do nothing. We would have to create these stub classes for every core control that might be associated with a database type, because the developer might want to default associations with those classes in this function. And the developer would have to do the same for his own custom controls. This is extra work for no functionality. 

The way I am suggesting it be done, PHP walks the inheritance chain and will pick the first one it finds. GetControlCodeGenerator would become:

```
public function GetControlCodeGenerator($objColumn) {
   $strClass = $this->ModelConnectorControlClass($objColumn);
   return $strClass ::GetControlGenerator();
}
```

Seems a lot cleaner to me, more flexible and a lot less effort to maintain. 

It seems like the hangup here is the worry over `$strClass`. The class name is already coming out of the ModelConnector dialog for some of the controls. There is no getting around it there, you have to use a string as a class name at some point.

spekary:
>This wouldn't achieve the main goal, which is keeping the core independent from the codegen.

I am curious why this is important to you? What are you trying to achieve? Do you have a particular project where this is important?

vakopian:
> > This wouldn't achieve the main goal, which is keeping the core independent from the codegen.

> I am curious why this is important to you? What are you trying to achieve? Do you have a particular project where this is important?

Until now the QCubed Core had no dependency to codegen. If I did not want codegen, I could have deleted the codegen directory, wrote all my code by hand, I would not even need to know that there is such a thing as codegen. In fact if I wanted to, I could have used a completely different code generator. BTW, this is not too theoretical, there are many other php code generators out there. At some point we may even want to switch to one of them instead of ours.

These dependency may seem small now, but it always starts small. You and I may agree that it's small. But the next developer comes in, and says "well, there is already one dependency, so why can't I add another one". And before you know it, you're looking at tight coupling.

Also, I'm not the only one who sees this dependency as a bad thing. In the discussion thread for this issue, pretty much every other involved QCubed developer agreed with this.

In this particular case we can avoid the dependency and still provide all the features you developed.

vakopian:
I think I have a good compromise solution. Checkout commit a8f6419. As you can see I brought back your old `ModelConnectorControlClass` (except it's now protected). In the new `GetControlCodeGenerator` I first check if the control class has a method called `GetCodeGenerator`. If it does, then that's what will be used. So you are free to introduce the coupled code in your own controls if you so wish. But the core controls do not have that method (so no dependency in the core). Instead the method returns the generator objects for the core controls on a case by case bases (using the actual object instantiate rather than based on class name). And at the end it falls back to my parent lookup solution as before.

So is this acceptable as a compromise?

vaibhav-kaushal:
I think the current solution is okay. Though in 3.1 I would like to bring the Code Generator out. I have some plans to extend the power of the code generator and for that, certain parts must stop their dependency on the Core controls. 

About other code generators - I do not think anything comes close to what we have got in QCubed. I like the integrated approach a lot more than others where the solution is either fragmented or they are wrapped into a Windows executable, are closed source and can cost up to US $499. They might be powerful and the price may be justified but I can't modify those at will.

And about the compromise - I think it is a good one.

spekary:
The compromise seems fine, but needs testing before committing it. I think we will still need `GetCodeGenerator()` functions in all the core controls, but it can wait until that is clear.

However, I still feel the need to defend the direction of this. 

QCubed has always had a certain amount of coupling to achieve its goals of being RAD and easy for new developers to use. Many frameworks are moving this direction. Node, Angular, Knockout, some Ruby frameworks, are designed to be able to easily create controls that are connected to database fields and give instant update and feedback. They achieve those results in a variety of ways. QCubed does it with code generation.

QCubed is not a modular framework, never has been, never has tried to be. There are other modular frameworks out there, like Symphony and Yii. If you want to take QCubed in that direction, you should fork it into a new project. QCubed is tightly integrated, and you will be hard-pressed to undo that.

Remember that we are currently only talking about QControls generating code for the ModelControllerBase (aka MetaControlBase). We are not placing limits on how a control is used. There is no "dependency" on code generation or the code generator. Nothing is preventing a developer from keeping the old templates and defining everything with the templates. The code generators never even get brought into PHP in the normal course of events, it only happens during code generation, and only if the templates ask for it. Everything is still controlled by the templates.

The benefits of having QControls generate their own "Controller" code are huge:

1) Easily create subclasses of controls that can also connect to database fields without needing to create custom template code.
2) On-the-fly changes of control types using the ModelController dialog.
3) Setting of many of the `__set` options for a control through the dialog. These options are customized based on the control type itself.
4) Easily make changes to controls and control options late in  the development process, since so many options can now be specified at the generated code layer.

These make it much easier for all developers, and especially new developers, to customize a form and create an application.

All of this requires a shift in thinking about what a QControl is. It is not just a thing that spits out some html. A QControl is a concept of a user interface item that gathers user information and processes it, and that also presents up-to-date information. The processing could be to place it in the database, or to hand it off to other code for further processing. A QControl has always included HTML, JS, and CSS generation. Version 3 adds PHP generation and watcher capabilities. 

There is a great danger in NOT coupling these things together. In the past, there have been many changes to controls that have broken the templates because the template code and QControl code are in completely different areas of the framework. Also, many developer customizations to a control have required creating custom templates that then get hard to maintain over time. 

The generated code in the ModelController is LOGICALLY connected to the control  and therefore should be physically connected to the control just for maintainability if nothing else. The previous complete lack of coupling between controls and code generation has caused a lot of headaches.

vaibhav-kaushal:
Where are we on this? Have we reached a conclusion?

vakopian:
> They achieve those results in a variety of ways. QCubed does it with code generation.

This is exactly my point - there are many ways to achieve this, and code generation is only one of them, and it may very well turn out not to be the best one at all. Thus, creating ***unnecessary*** dependencies to something that really is an implementation details is not the right approach.

I'm sure we can debate this even more, but at this point I think we both explained our points enough for anyone else to read and make up their minds. And I think the current compromise is pretty good - it allows us to be very close to your idea, but still keep things decoupled and modular.

vaibhav-kaushal:
IMHO, One of the most salient features of QCubed is the awesome code generation we have at hand. (The other one is automated Form Handling/Rendering mechanism). As such, QCubed will have to support code generation as perfectly as possible. Decoupling is not the problem - losing track of changes caused by the other part in the control-template duo is.

vaibhav-kaushal:
By the way, are we okay with this decision? Is it possible to at least move to beta?

vaibhav-kaushal:
@spekary Do we have a green flag here?

spekary:
Give me a bit to test this. The changes are extensive.

spekary:
We are almost at beta. There are some very minor tweaks I would like to do to this. And the plugins need to be updated to work with this, but we are feature complete at this point I think.

PR #659, opened 2015-03-16
bugfix-MYSQLI_ON_UPDATE_NOW

spekary:
Fix problem with flag. Defining the undocumented flag that detects whether a field is set to ON UPDATE CURRENT_TIMESTAMP

### DIFF
--- a/includes/database/QMySqliDatabase.class.php
+++ b/includes/database/QMySqliDatabase.class.php
@@ -1,4 +1,8 @@
 <?php
+
+if (!defined('MYSQLI_ON_UPDATE_NOW_FLAG')) {
+	define ('MYSQLI_ON_UPDATE_NOW_FLAG', 8192);
+}
 	/**
 	 *
 	 * @package DatabaseAdapters


includes/database/QMySqliDatabase.class.php
```diff
@@ -1,4 +1,8 @@
 <?php
+
+if (!defined('MYSQLI_ON_UPDATE_NOW_FLAG')) {
+	define ('MYSQLI_ON_UPDATE_NOW_FLAG', 8192);
```
vakopian: Is this really safe? If php (or the mysql driver) really added then deleted this flag, maybe this value is reserved for something else in the future. There is basically no way for us to know if the `8192` value does what we think it does.
Isn't it better to do something like this around the lines where the value is used:
<pre>
if (defined(MYSQLI_ON_UPDATE_NOW_FLAG)) {
	// Special situation that we take advantage of to automatically implement optimistic locking
	if ($mixFieldData->type == MYSQLI_TYPE_TIMESTAMP &&
			$mixFieldData->flags & MYSQLI_ON_UPDATE_NOW_FLAG) {
		$this->strType = QDatabaseFieldType::VarChar;
		$this->blnTimestamp = true;
	}
}
</pre>

spekary: I think in this case its safe. There is another way to do this. phpadmin parses the line definition clause looking for the on update current_timestamp phrase. However, since its in PHP Storm, my guess is someone took a look at the MySQL source and figured this out. If they did remove it, that would be a pretty big change for mysqli, even if it was only internal. If it does break, then it would mean collision detection would not work as expected, which is generally kind of a poor way of doing it anyway, so not a big loss.

PR #662, opened 2015-03-18
Updating the type handling of 'timstamp without time zone' data type in PostgreSQL database

vaibhav-kaushal:
Currently the data type is treated as VarChar while it should be treated as Date (string vs QDateTime types). This update fixes the behaviour.


vaibhav-kaushal:
I just changed the line numbers! Was trying to improve upon the behavior in 2.2 and later realise it has been already done in alpha. My bad.

PR #665, opened 2015-03-27
bugfix-ControlActions

spekary:
Need to clear control actions after rendering them.


includes/base_controls/QControlBase.class.php
```diff
@@ -1209,6 +1209,7 @@ public function RenderAttributeScripts()
 					call_user_func_array('QApplication::ExecuteControlCommand', $scriptArgs);
 				}
             }
+			unset ($this->strAttributeScripts);
```
vaibhav-kaushal: Why not set it to null?

PR #676, opened 2015-04-18
bugfix-ConnectorClassName

spekary:
Making this public so that it is usable by the code generation routines to get the name of the current class. Its a little faster and easier than GetCodeGenerator()->GetClassName()


vakopian:
The reason that method is protected is precisely so that it is not used in the code generation routines. The whole point of the `AbstractControl_CodeGenerator` class and its hierarchy is to define the interface by which the code generation access these methods. If that method is called directly by the code generator, then subclasses of `AbstractControl_CodeGenerator` have no way to override the behavior.

PR #681, opened 2015-05-14
bugfix-JsCommandOrderAndRestoreState

spekary:
Rearranging where javascript commands output so that they can be executed during form create. Fixing a problem that wherbey if you set an option in a control that needed to execute a javascript command to set something in the corresponding html widget, it would not work because the widget was not yet initialized.

Also fixed a redundant restoring of control state. Control states are already restored individually by setting SaveState to true.

### DIFF
--- a/includes/base_controls/QFormBase.class.php
+++ b/includes/base_controls/QFormBase.class.php
@@ -543,9 +543,6 @@
 				// Trigger Create Event (if applicable)
 				$objClass->Form_Create();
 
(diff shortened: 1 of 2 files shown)


spekary:
As I thought, this is breaking other things. The .off in the some of the jQueryUI widgets (like dialogs) is now happening after the events are attached to the objects, so those events aren't firing. Will have to rethink this.

Issue #690, opened 2015-05-20
Installer step 3 "Launch the config checker" link has no host if either variable in the link is empty


Elyrith:
If either $strVirtDir or $strSubDir are empty then the "Launch the config checker" link doesn't include the host and path. The link is exactly "http://assets/_core/php/_devtools/config_checker.php". For some reason if even one of those variables is NOT empty then there are no problems.

How to replicate:
1. Create your web root at /home/qcubeduser/qcubed/
2. Download the tarball of 2.2.3 from https://github.com/qcubed/framework/releases
3. Put QCubed content in the root (not in a subfolder or virtual directory)
4. Run through the installer to step 3 and check/click the "Launch the config checker" link

I have a fix that I'll create a pull request for in a few minutes.

Elyrith:
Now that I'm looking at things I've noticed this problem in a couple of different places in odd situations, but not in others. I'm going to close the pull request and test a few more things.

For example, if I pull down QCubed 2.2.3 directly from GitHub with no virtual directories or subdirectories (accessible for me at http://qcubed.example.com) the stylesheet and images are broken in the same way (http://assets/_core/css/styles.css and http://assets/_core/images/qcubed_logo_footer.png respectively). If I put everything in a folder (http://qcubed.example.com/qcubed) then the stylesheet, images, and "Launch the config checker" link all work perfectly.

These results are all on Ubuntu 14.04 with Apache 2.4.7 and PHP 5.5.9.

PR #691, opened 2015-05-20
Don't add the strVirtDir and strSubDir variables to the Launch link if they are both empty.

Elyrith:
This is a patch to fix issue #690. If there is no content in strVirtDir or strSubDir then don't add them to the link path since it removes the host.

**Please, please, please test these before you merge.** I tested it 5 times on Ubuntu 14.04 on DigitalOcean with Apache 2.4.7 and PHP 5.5.9:

Three tests were putting it in different locations on the file system but always in the virtualhost root: Always had the issue from #690, but this patch fixed them all.

One test was with a virtual directory and I didn't have any problems with the link before or after the patch was applied.

The last test was with a subdirectory and I didn't have any problems with the link before or after the patch was applied.

I used existing code as examples for style, so I'm pretty sure that's okay, but let me know if there's anything else that needs fixing, or if it doesn't work for you, or if you can't replicate the issue, or... anything.

### DIFF
--- a/assets/_core/php/_devtools/installer/step_3.php
+++ b/assets/_core/php/_devtools/installer/step_3.php
@@ -183,7 +183,11 @@
 				switch ($strFileCreationStatus) {
 					case 'exists':
 					case 'creation_failed':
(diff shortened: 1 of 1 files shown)


Elyrith:
Closing temporarily(?) for more testing. See issues #690 for more info.

Issue #692, opened 2015-05-22
QSqlServer2005Database.class.php - Codegen fails if the table has no index column and no primary key


ubs:
I encountered this issue while trying to debug a connection to Microsoft SQL Server DB.
### Issue

When a table has no indexes, the code gen fails with an unclear error. It turns out that it fails when an undefined function is called to get the error information from the DB resource. This rogue piece of code can be found in 

`QSqlServer2005Database.class.php line 553 (QSqlServer2005DatabaseResult->FetchArray())` 

I believe this is a copied and pasted code as the `GetErrorInformation()` method does not exist within the scope specified.
### Resolution
1. Solve the immediate issue of undefined `GetErrorInformation()` method by slightly altering the line as follows:

`$this->objDb->GetErrorInformation($strErrorinformation, $strErrorCode);`

Also, the variable `$strNonQuery` does not exist on line 554 so I altered that as follows:

`throw new QSqlServer2005DatabaseException($strErrorinformation, $strErrorCode, null);`
1. Solve the NO INDEX problem pre-emptively.
   As it turns out, Qcubed requires at least one index (e.g. the primary key) in order to successfully generate the ORMs. I added additional code to `QSqlServer2005Database.class.php` to check if the table being analysed has at least one index. If not, it should fail gracefully or NOT, as you may decide it to be. 

See lines 440 - 496 of attached code file `QSqlServer2005Database.ProfNoTime.class.php`

I hope this gets corrected in the latest branch/version.

vaibhav-kaushal:
You can make a pull request for the same :)

ubs:
Done. Pull request created.

PR #693, opened 2015-05-22
Pull Request for - QSqlServer2005Database.class.php - Codegen fails if the table has no index column and no primary key (Issue #692)

ubs:
Please see link below for issue description.

[Issue 692 - Description Link](https://github.com/qcubed/framework/issues/692)

### DIFF
--- a/includes/qcubed/_core/database/QSqlServer2005Database.class.php
+++ b/includes/qcubed/_core/database/QSqlServer2005Database.class.php
@@ -438,6 +438,19 @@
 
 		public function GetIndexesForTable($strTableName) {
 			$objIndexArray = array();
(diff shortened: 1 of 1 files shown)


vaibhav-kaushal:
Does this pull request actually work? (I do not have Microsoft SQL server installation to verify).

spekary:
Don't need to test every pull request yourself. Our policy has been to do a code review, and if what is desired is correct and the code looks like its going to accomplish its mission without harming other code, we commit it. If someone has a problem with it later, it will come back. Its part of being in continuous integration mode.

Its too bad that travis doesn't support SqlServer, but that can't be helped.

PR #701, opened 2015-06-04
Added a few useful properties to QDateTimePicker

vaibhav-kaushal:
Added the AllowBlankTime, AllowBlankDate, TimeSeparator, SecondInterval, MinuteInterval and HourInterval property.

**AllowBlankTime** and **AllowBlankDate** can be used to remove the ```--``` entry from the DateTimePicker. This is helpful where we have to force the user to enter valid time. 

**TimeSeparator** allows you to select the character used for separate time.

**HourInterval**, **MinuteInterval** and **SecondInterval** allow the developer to select intervals that he wants to let the user select from (e.g. I want to allow selection of time only in intervals of 10 minutes).

These properties allow us to escape a lot of validation logic at times when extreme strictness is not required but certain rules (which are obvious by these properties) are helpful.

### DIFF
--- a/includes/base_controls/QDateTimePicker.class.php
+++ b/includes/base_controls/QDateTimePicker.class.php
@@ -267,7 +267,7 @@
 					if (!$this->blnRequired || $dttDateTime->IsTimeNull())
 						if ($this->blnAllowBlankTime)
 							$strHourListBox .= '<option value="">--</option>';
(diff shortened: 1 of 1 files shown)


vaibhav-kaushal:
I have added this and have been working on these since 2.2 without problems. So far with all my tests, 3.0 has not given any trouble either. In case no one else tests this, I will merge it in coming couple of days.

PR #703, opened 2015-06-09
Restoring DeleteCache method to the generated ORM classes.

vaibhav-kaushal:
In the current Beta 3.0 branch, the templates do not have the DeleteCache method which makes the programs fail. This pull request is to bring it back.


spekary:
This has been renamed to DeleteFromCache. See the QModelTrait.trait.php file.

PR #714, opened 2015-06-28
Update INSTALL.md

matthiaz:
Minor typo correction


spekary:
I am going to merge to beta3 and then take down alpha3

vaibhav-kaushal:
That would be better! I did not check the target branch before accepting!

PR #725, opened 2015-07-08
bugfix-PluginManager

spekary:
Completing the removal of the old installer. 

Commented out old code instead of removing it. Potentially the installer could be reworked to directly call the composer command, but its not so easy because people can set up composer in different ways. We somehow have to capture how the user executes the composer command in order to automate the process. Or, we call into github using the github api and bypass composer.


kukrik:
@spekary! Please to merge 1 commit into qcubed:beta-3.0 from spekary:bugfix-PluginManager.
It would be necessary!

Issue #764, opened 2015-08-18
Model class paths for autoload missing?


vaibhav-kaushal:
I installed QCubed 3.0 using https://github.com/qcubed/app-starter. It worked fine (except that I had to set the values of ```__SUBDIRECTORY__``` and ```__DOCROOT__``` manually for them to be correct).

However, after code-generation, my model classes are not loading. What might be the reason? I could not find QQN.class.php either :confused: .

spekary:
Something has caused the templates to fail in the middle. Will need some debugging.

vaibhav-kaushal:
I think I have found out the problem. A little bit more debugging is required. I think it is because we are serving warnings and errors both as errors and then the if condition here in https://github.com/qcubed/framework/blob/beta-3.0/assets/php/_devtools/codegen.php

```
			if (!$strErrors) {
				foreach (QCodeGen::GenerateAggregate() as $strMessage) { ?>
					<p><strong><?php _p($strMessage); ?></strong></p>
				<?php }
			}
```
causes the trouble. My schema has some foreign keys without single column index on them. Hence I get some warnings (which are all set as errors) and so, the aggregator part never runs.

vaibhav-kaushal:
This issue was fixed. Closing thread. Please see #765 for solution.

PR #765, opened 2015-08-22
Split warnings and errors generated in the code generation system

vaibhav-kaushal:
Splitting warning and errors in the code-generation process to ensure that the aggregation system works if there are only warnings, not errors.


vaibhav-kaushal:
Solves the problem mentioned in #764 .

PR #772, opened 2015-08-31
feature-MysqlStoredProcedure

spekary:
Added function to execute a Mysql stored procedure. Could be generalized to other DBs.


spekary:
@olegabr, can you add the Postgresql version of this?

PR #773, opened 2015-09-01
Added some documentation

vaibhav-kaushal:
Added some documentation. No code has been altered.


spekary:
What is your opinion on whether we need @access tags in the comments?

vaibhav-kaushal:
I think phpDocumentor handles it automatically. Also, IDEs can resolve the autocomplete (and related errors) quite easily. Adding that isn't required, really. The less the number of tags, the cleaner it looks. Also, adding that would require significant work.

PR #777, opened 2015-09-02
bugfix-DecimalType

spekary:
Making Decimal values be PHP strings instead of floats. The whole purpose for decimals is precise tracking of fractions, and converting to floats completely looses that benefit. It also prevents using the bcmath library to do precision math with the Decimal number. If the developer wants a float, they should  just specify a float. Anyways, if the developer still wants a float, PHP will do the conversion automatically when it is used in a math operation.

### DIFF
--- a/includes/database/QMySqli5Database.class.php
+++ b/includes/database/QMySqli5Database.class.php
@@ -130,7 +130,7 @@
 		protected function SetFieldType($intMySqlFieldType) {
 			switch ($intMySqlFieldType) {
 				case MYSQLI_TYPE_NEWDECIMAL:
-					$this->strType = QDatabaseFieldType::Float;
+					$this->strType = QDatabaseFieldType::VarChar;
 					break;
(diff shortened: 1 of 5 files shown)


vaibhav-kaushal:
Are you sure about this? What happens when we try to save the variables back into the database? Aren't we going to face the problem when casting inside ```SqlVariable``` method when ```Save()``` is called in the generated ORM class file?

spekary:
Should not be a problem. If the developer leaves it as a string, that is what SQL expects as input for DECIMAL values. If the developer converts to a float, then the result will be what is currently happening in our code base. The float value will be printed in the SQL statement without quotes, and SQL knows how to convert a FLOAT to a DECIMAL. Essentially this change gives the developer the option of maintaining strict precision by only working with strings. But the developer can always just ignore that and use a string as a float, and PHP will convert it to a float at that moment.

PR #794, opened 2015-10-26
bugfix-PaginatorVariable

spekary:
This variable is defined in the subclass only, but used in the superclass. This fix adds the definition to the superclass too.


vaibhav-kaushal:
This is one change I had done on my end and was going to create a new pull request for it.

PR #795, opened 2015-10-26
bugfix-EmptyDataRepeater

spekary:
Fixing problem with data repeater. If the data binder returns no data, the data repeater is removed from the dom. If the data binder then returns data later, the data grid will be placed back in the dom, but at the end of the form since there is no remnant of the data repeater to know where to put it.

### DIFF
--- a/includes/base_controls/QDataRepeater.class.php
+++ b/includes/base_controls/QDataRepeater.class.php
@@ -160,13 +160,13 @@ class QDataRepeater extends QPaginatedControl {
 				$this->intCurrentItemIndex++;
 			}
 
-			$strToReturn = $this->RenderTag($this->strTagName,
-				null,
-				null,
-				$strEvalledItems);
-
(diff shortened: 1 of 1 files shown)


vaibhav-kaushal:
I do not understand the need of this one. Can you elaborate? :confused:

spekary:
This is a bug fix for ajax drawing of a data repeater that has no wrapper. Data repeater's use data binders. Data binders often depend on what is going on in other controls in the form. If the user creates a situation where the data binder returns no data, the old code was removing the entire control from the dom, including the beginning and ending div tags. If the user then changes something, the control gets sent to the browser through the javascript ajax listener. The problem is that once the control is removed, there is no way to tell where to put it back, and so it gets added to the end of the form, rather than where it was when it was first created. This fix makes sure the beginning and ending div tags are still in the dom, even if no data is drawn inside them, so that when data DOES get drawn, the ajax javascript code knows where to put it.

PR #797, opened 2015-10-26
feature-OverrideButtonDraw

spekary:
Allowing inner-html of button to be over-ridden. Useful for bootstrap buttons.


vaibhav-kaushal:
That's a nice little change :) Thanks.

PR #799, opened 2015-10-26
bugfix-ForceBlockElement

spekary:
Enables the support for forcing an element to be drawn as a block-level element if it has a wrapper. Also permits defining render methods with additional paramenters that are not related to overriding parameters.


vaibhav-kaushal:
Line 1269 is showing it red (beings with a back slash ```\```). Except that everything is okay!

PR #812, opened 2015-11-14
Updated connectors example to work with new QDatagrid2 architecture

koopa:


spekary:
Thank you for contributing!

PR #824, opened 2015-11-30
bugfix-Multilne

spekary:
Fixing problem with submits of multi line text boxes. Extra lines were being added. This was caused by switching to HTML5 encoding. The fix is to convert \r\n combinations in post variables to a single newline.

This also fixes a potential security problem. Warnings and errors were being sent to the browser in the clear without encoding the text.


vaibhav-kaushal:
My system is also a Mac with Chrome and it is sending ```\r\n```.

vaibhav-kaushal:
It is working as expected.

PR #855, opened 2015-12-24
Fixes the app-starter issue

olegabr:
https://github.com/qcubed/app-starter/issues/10

Seems that the `_p` function's default behavior has changed.


olegabr:
Installer complete now. I see no more issues with it.

Issue #856, opened 2015-12-24
Codegen complains for new project


olegabr:
Codegen complains that:
```
Unable to mkdir /var/www/example.com/project/forms
Unable to mkdir /var/www/example.com/project/includes/panel
Unable to mkdir /var/www/example.com/project/includes/dialog
```

These directories should pre-exist in an install/project/ directory with write rights requested in composer file, and `assets/php/_devtools/config_checker.php` should check it's existance and write rights.

olegabr:
fixed in #857

PR #857, opened 2015-12-24
Fixes the #856 issue

olegabr:
Codegen complains for new project

Codegen complains that:
```
Unable to mkdir /var/www/example.com/project/forms
Unable to mkdir /var/www/example.com/project/includes/panel
Unable to mkdir /var/www/example.com/project/includes/dialog
```

These directories should pre-exist in an install/project/ directory with write rights requested in composer file, and `assets/php/_devtools/config_checker.php` should check it's existance and write rights.


olegabr:
Also note changes in the `assets/php/_devtools/codegen.php` file. It makes all `Successfully generated DB ORM Class` sentences to be on the same padding. Without this change the codegen results looks unprofessional.

Issue #861, opened 2015-12-24
Permission denied while Save and Regenerate button is pressed


olegabr:
```
file_put_contents(/var/www/example.com/project/includes/configuration/codegen_options.json): failed to open stream: Permission denied
```

1. composer.json should require it to be writable
2. config checker should check it to be writable

olegabr:
fixed in #863

PR #863, opened 2015-12-25
Fixes the #861 issue (Permission denied while Save and Regenerate button is pressed)

olegabr:
1. The 'empty' `codegen_options.json` with `{}` content is created
 2. composer.json require it to be writable
 3. config checker check it to exists and be writable


spekary:
Looks like other pull requests conflict here. Can you reissue? I thing this is a reasonable fix. Oh, and good to see you back!

olegabr:
conflicts are resolved now

Issue #869, opened 2015-12-26
Examples errors


olegabr:
- [X] `assets/php/examples/code_generator/intro.php` The picture of the db structure is outdated
 - [X] `profile.php` in the value of field `Profile Generated From` HtmlEntities function scramble text making it human unreadable.
 - [X] `assets/php/examples/more_codegen/custom.php` `Undefined GET property or variable in 'Project' class: fltBudget` It is becase Budget is generated with type string, not float. In the `assets/php/examples/mysql_innodb.sql` file it has type `DECIMAL`: `budget DECIMAL`
 - [X] `assets/php/examples/more_codegen/optimistic_locking.php` `Attempting to save a Two Instances of the Same Object will throw an Optimistic Locking Exception` if this button is pressed, error page is shown. Alert box would be helpfull here:
```
Exception in PHP Script /example.com/vendor/qcubed/framework/assets/php/examples/more_codegen/optimistic_locking.php
Optimistic Locking constraint when trying to update PersonWithLock object.  To update anyway, call ->Save() with $blnForceUpdate set to true

Exception Type: QOptimisticLockingException
```

 - [X] `assets/php/examples/qcubed_query/qcache.php` Press the `Clear the cache` button. Message is shown: `Clearing the query cache for the Person table was NOT successful - check your cache / namespace paths. Reload the page to see the effect - query will not be executed against the database.`
 - [X] On any example page in the generated code:
```html
		<link rel="stylesheet" type="text/css" href="&sol;example&period;com&sol;vendor&sol;qcubed&sol;framework&sol;assets&sol;css&sol;styles&period;css"></link>
		<link rel="stylesheet" type="text/css" href="&sol;example&period;com&sol;vendor&sol;qcubed&sol;framework&sol;assets&sol;php&sol;examples&sol;includes&sol;examples&period;css"></link>

		<div id="tagline"><a href="http://qcubed.github.com/" title="QCubed Homepage"><img id="logo" src="&sol;example&period;com&sol;vendor&sol;qcubed&sol;framework&sol;assets&sol;images&sol;qcubed&lowbar;logo&lowbar;footer&period;png" alt="QCubed Framework" /> <span class="version">3&period;0&period;0 Beta Release &lpar;QCubed 3&period;0&period;0&rpar;</span></a></div>
```
It is the same problem with `_p` in non-human readable parts like href or src.
[...]

PR #870, opened 2015-12-27
Issue #869 (Examples errors)

olegabr:
- `profile.php` in the value of field `Profile Generated From` HtmlEntities function scramble text making it human unreadable.
 - On any example page in the generated code:
```html
		<link rel="stylesheet" type="text/css" href="&sol;example&period;com&sol;vendor&sol;qcubed&sol;framework&sol;assets&sol;css&sol;styles&period;css"></link>
		<link rel="stylesheet" type="text/css" href="&sol;example&period;com&sol;vendor&sol;qcubed&sol;framework&sol;assets&sol;php&sol;examples&sol;includes&sol;examples&period;css"></link>

		<div id="tagline"><a href="http://qcubed.github.com/" title="QCubed Homepage"><img id="logo" src="&sol;example&period;com&sol;vendor&sol;qcubed&sol;framework&sol;assets&sol;images&sol;qcubed&lowbar;logo&lowbar;footer&period;png" alt="QCubed Framework" /> <span class="version">3&period;0&period;0 Beta Release &lpar;QCubed 3&period;0&period;0&rpar;</span></a></div>
```
It is the same problem with `_p` in non-human readable parts like href or src.


olegabr:
Please, do not merge it yet, I'm going to add commits here.

olegabr:
I've resolved 9 of 22 points for now. Fill free to work on remaining ones )

assets/php/examples/events_actions/javascript_alerts.tpl.php
```diff
@@ -17,7 +17,12 @@
 	javascript function on the page itself, so that the button has some javascript to perform.</p>
 	
 	<p>If you are interested in more advanced and flexible types of confirmation or prompts, take a look at the 
-	<a href="../../../../plugins/DialogBoxes/example/more_dialog_boxes.php">plugin that offers pre-built QDialogBox subclasses</a>.</p>
+		<?php if (file_exists('../../../../plugins/DialogBoxes/example/more_dialog_boxes.php')) { ?>
+		<a href="../../../../plugins/DialogBoxes/example/more_dialog_boxes.php">plugin that offers pre-built QDialogBox subclasses</a>.
```
spekary: This is not a 3.0 plugin, and some dialog functionality has been added to 3.0. I will commit this, but we need to look at this again before releasing.

assets/php/examples/master_detail/project_list.php
```diff
@@ -10,9 +10,12 @@ class ProjectListForm extends QForm {
...
 					
 		protected function Form_Create() {
-			// Instantiate the Meta DataGrid
 			//$this->objDefaultWaitIcon = new QWaitIcon($this);
-			$this->dtgProjects = new ProjectDataGrid($this);
+
+			// Instantiate the DataGrid
+			// It is a simple QDataGrid because the generated class ProjectList
+			// is inherited from the QDataGrid2 now
+			$this->dtgProjects = new QDataGrid($this);
```
spekary: QDataGrid is obsolete. Why are we moving backwards here?

assets/php/examples/master_detail/project_list.php
```diff
@@ -49,33 +52,38 @@ protected function Form_Create() {
...
 			// remove filter from ID column
-			$colId = $this->dtgProjects->AddConnectedColumn('Id');
+			$colId = new QDataGridColumn('Id', '<?= $_ITEM->Id ?>');
 			$colId->FilterType = "";
+			$this->dtgProjects->AddColumn($colId);
 
-			$this->dtgProjects->AddConnectedColumn('Name', 'Name=Project');
-			$this->dtgProjects->AddDbTypeColumn('ProjectStatusTypeId', 'ProjectStatusType');
+			$this->dtgProjects->AddColumn(new QDataGridColumn('Project', '<?= $_ITEM->Name ?>'));
+			$this->MetaAddTypeColumn(QQN::Project()->ProjectStatusTypeId, 'ProjectStatusType');
```
spekary: MetaAddTypeColumn is deprecated.

assets/php/examples/master_detail/project_list.php
olegabr: I've tried to move to the QDataGrid2 here, but it does not support all QDataGrid features used in this example. It is quite complicated with datagrid in datagrid in dtagrid )

assets/php/examples/master_detail/project_list.php
olegabr: The same reason here - I don't know how this whole example can be implemented in the QDataGrid2 and if it is possible in the first place. Moreover, if QDataGrid is deprecated there are many places where it is used in examples. Should we replace it in all places?

spekary:
Can you pull out the QDataGrid fixes and put them in a separate issue, and I will take a look? This has become a hard push request to commit because too much is changing.

olegabr:
@spekary, It is ready to be merged!

olegabr:
Sorry, missed your comment

olegabr:
Do you want me to revert the https://github.com/olegabr/framework/commit/ee7548cc594a84219e262c613796182b633817a8 commit here for now?

olegabr:
The revert is done. The issue moved to the #881.

olegabr:
@spekary, are you planning to take a look on tis pull request, or I can merge it myself?

install/project/includes/configuration/configuration.inc.sample.php
```diff
@@ -165,9 +166,11 @@
 			define ('__APP_CSS_ASSETS__', __PROJECT_ASSETS__ . '/css');
 			define ('__APP_IMAGE_ASSETS__', __PROJECT_ASSETS__ . '/images');
 			define ('__APP_PHP_ASSETS__', __PROJECT_ASSETS__ . '/php');
+			define ('__APP_IMAGE_CACHE__', __APP_IMAGE_ASSETS__ . '/cache');
+			define ('__APP_UPLOAD__', __DOCROOT__ . __PROJECT_ASSETS__ . '/upload');
```
spekary: IMHO caches and uploads should be relative to the __TMP__ directory by default.

install/project/includes/configuration/configuration.inc.sample.php
```diff
@@ -165,9 +166,11 @@
...
 			
 			define ('__PLUGIN_ASSETS__',  __SUBDIRECTORY__ . '/vendor/qcubed/plugin');
-			define ('__IMAGE_CACHE__', __APP_IMAGE_ASSETS__ . '/cache');
+			define ('__IMAGE_CACHE__', __IMAGE_ASSETS__ . '/cache');
```
spekary: This too should be relative to the __TMP__ directory. __IMAGE_ASSETS__ is in the vendor directory, which is controlled by composer

spekary:
Almost there. I made a comment about having uploads and caches go to the ```__TMP__``` directory by default.

install/project/includes/configuration/configuration.inc.sample.php
kukrik: We must then distinguish between that which the assets or the temporary folder (TMP). I do not like a big mess, this is too much and cache folders everywhere. This is not good. Here you can agree on that, whether it is sufficient for the development or production of, for example, from two places:
- project/assets (permanent stuff)
  - uploads
  - images
  - cache
  - i18n.po
- project/tmp (temporary things that could, if necessary, to put Cron periodic cleaning, etc.)
  - purifier
  - soap
  - etc

It is easier to track and maintain so. What do your thinks?

spekary: I am concerned about security. Anything that can be uploaded by the user should be treated carefully, and putting it in the project/assets directory is not careful. However, we are uploading things that need to be directly accessed by the browser, so the TMP directory is not a good place for that. Not sure how to handle that.

olegabr: As @kukrik mentioned already, we need two cache folders: one for php-accessible things, and it should be full filesystem path with `__DOCROOT__`, and another one for web-accesible goods, and it should be relative to site domen, e.g. for `http://example.com/project/assets/cache/myphoto.png` it should be `/project/assets/cache`

For the first one we have `__TMP__`. For the second we don't have one place, but only images cache in the `__APP_IMAGE_CACHE__` define. We should have `__APP_CACHE_ASSETS__ `instead and, may be, `__APP_IMAGE_CACHE_ASSETS__` defined as `__APP_CACHE_ASSETS__  . '/images'`. And we need to have `__APP_CACHE__` as a full-path version of the  `__APP_CACHE_ASSETS__` with a `__DOCROOT__` included. The `__APP_IMAGE_CACHE__` becomes a full-path version of the `__APP_IMAGE_CACHE_ASSETS__` define.

Uploads are different. All other caches a volatile and can be cleared in any time. Uploads are not. They are a part of user expirience and should be preserved. That is a reason to maintain it separately from cache directory. It means that the current `__APP_UPLOAD__` is done right. But it lacks it's `ASSETS` counterpart: `__APP_UPLOAD_ASSETS__` that is needed to be used by client code to form URI-s of uploaded items. The original `__APP_UPLOAD__` is used by the code that needs to work with file in a filesystem.

If it is accepted, I can implement these changes.

olegabr:
Great!

It looks like we have 20 issues for v3 here: https://github.com/qcubed/framework/milestones/3.0%20Release

install/project/includes/configuration/configuration.inc.sample.php
kukrik: Hmmm, @spekary, you're right, I forgot that security is very important. I'm very sorry. The principle is that it is not good that the folders are a several of places.

Folders to organize and should be separated from this principle, for the function of which is intended for one or the other. It is important only to, at least in one place.

PS. You should definitely take advantage of these folders configuration.inc.php comment.

PR #874, opened 2016-01-09
bugfix-TreeNavIssues

spekary:
Fixes to handle the unique way that TreeNav delivers ajax events. Allows other potential controls to be able to directly send ajax events and then look for the result in the ParsePostData function.

### DIFF
--- a/includes/base_controls/QFormBase.class.php
+++ b/includes/base_controls/QFormBase.class.php
@@ -479,12 +479,18 @@
 				}
 				else {
 					// Ajax post. Only send data to controls specified in the post to save time.
+					$previouslyFoundArray = array();
 					foreach ($_POST as $key=>$val) {
-						$strControlId = $key;
+						if ($key == 'Qform__FormControl') {
+							$strControlId = $val;
+						} elseif (substr($key, 0, 6) == 'Qform_') {
+							continue;	// ignore this form data
+						} else {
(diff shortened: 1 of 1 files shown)


includes/base_controls/QFormBase.class.php
```diff
@@ -479,12 +479,18 @@ public static function Run($strFormId, $strAlternateHtmlFile = null) {
 				}
 				else {
 					// Ajax post. Only send data to controls specified in the post to save time.
+					$previouslyFoundArray = array();
 					foreach ($_POST as $key=>$val) {
-						$strControlId = $key;
+						if ($key == 'Qform__FormControl') {
+							$strControlId == $val;
```
olegabr: double `==` is a mistake here

includes/base_controls/QControlBase.class.php
```diff
@@ -257,11 +257,16 @@ public function __construct($objParentObject, $strControlId = null) {
 			if (strlen($strControlId) == 0)
 				$this->strControlId = $this->objForm->GenerateControlId();
 			else {
+				if (substr($strControlId, 0, 6) === 'Qform_') {
```
olegabr: `ctype_alnum` already ensures that no underscore symbol would be allowed. This check is unnecessary.

PR #877, opened 2016-01-10
test and example for issue #439 Expand doesn't take Aliases into account

olegabr:
- It is the same as the PR #441, but for beta-3.0 branch
 - The original fix #440 was wrong, but in beta-3.0 it is completely outdated.
 - This PR is to be sure that the original problem is not reproduced in the new code of v3.0


olegabr:
ready to be merged. do not forget to close #439 as well

PR #895, opened 2016-01-17
bugfix-QcodoRemoval

spekary:
Removing as many references as possible of 'qcodo'. There are still a few left based on the context.


assets/php/examples/communication/rss_feed.php
```diff
@@ -1,15 +1,15 @@
 <?php require_once('../qcubed.inc.php');
 
 	// Setup the Feed, itself
-	$objRss = new QRssFeed('Examples Site Projects', 'http://examples.qcodo.com/', 'An Example RSS feed of the Qcodo Examples Site Projects');
-	$objRss->Image = new QRssImage('http://www.qcodo.com/images/qcodo_smaller.png');
+	$objRss = new QRssFeed('Examples Site Projects', 'http://examples.qcu.be/', 'An Example RSS feed of the Qcubed Examples Site Projects');
+	$objRss->Image = new QRssImage('http://www.qcu.be/images/qcodo_smaller.png');
```
olegabr: qcodo still here )

spekary: Yep. I have no idea if this stuff is even on the qcu.be server, but if it is, I bet the file name hasn't changed. We can fix this one later.

PR #930, opened 2016-03-13
feature-SetCookie

spekary:
Set cookie function that works in both Ajax mode and Server mode.


kukrik:
@spekary! Good to hear! How would a session? I think of it like that...
The session is safe, and it goes to the database also need to occasionally propel and consultation, etc.

What do you think?

spekary:
Sorry, your translation doesn't make sense to me.

If you are asking about session variables, they should already work in ajax mode.

kukrik:
Sorry, thanks!

PR #934, opened 2016-03-16
feature-QSimpleTableLinkColumn

spekary:
A multi-purpose column to display anchor tag links. An answer to the "Edit" column that is now deprecated from QDataGrid. This column is much more flexible, but still very easy to use. Can accommodate  static text, property chains, multi-dimensional arrays,  and callbacks to use as the label, the destination, the GET variables, or the anchor tag attributes. Can display a QControlProxy. Can be configured to display a minimal link such that a QOnEvent can be used to trap the bubbled up click if needed.

### DIFF
--- a/includes/base_controls/QSimpleTableBase.class.php
+++ b/includes/base_controls/QSimpleTableBase.class.php
@@ -161,12 +161,6 @@
 		/**
 		 * Add a link column.
 		 *
-		 * @param $strName
-		 * @param $strAttribute
-		 * @param $intColumnIndex
-		 * @return QVirtualAttributeColumn
-		 */
-		/**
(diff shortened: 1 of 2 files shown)


includes/base_controls/QSimpleTableBase.class.php
```diff
@@ -158,6 +159,50 @@ public function CreateVirtualAttributeColumn ($strName, $strAttribute, $intColum
 		}
 
 		/**
+		 * Add a link column.
+		 *
+		 * @param $strName
```
olegabr: looks like auto-generated stubs are mistakenly added

spekary: Doh.

PR #939, opened 2016-03-21
Database connection encoding

kukrik:
Umlauts problem is always a problem! In English language not problem.
What do you think of this?


kukrik:
My mistake, I would like to present just one pull "Proposal to add encoding "UTF-8" Database Connection". I asked to pull all of that is not necessary, sorry...

Do I must do it again?

PR #942, opened 2016-03-21
Proposal to add encoding "UTF-8" to Database Connection

kukrik:
Umlauts problem is always a problem! In English language not problem.
This correction allows umlauts in the database to display correctly.

What do you think of this?


kukrik:
Close again. Sorry again!

PR #945, opened 2016-03-30
RenderAsScript function added

matthiaz:
If you use RenderAsHref you get onclick="qc.Pa... "; but if you want the raw javascript, you have to create this yourself.

I created a new function RenderAsScript which allows you to get the raw javascript action:
qc.Pa('FrontController','c62','QClickEvent',....


spekary:
That is fine, but the QControlProxy mechanism has been significantly rewritten in v3 and works on data attributes rather than embedded scripts. If you use the function you just created, it will not be compatible with beta 3.

matthiaz:
@spekary I took a look at qcubed beta to see if this could be ported. Turns out it isn't too hard to port. 
I made a pull request for it here: https://github.com/qcubed/framework/pull/946

PR #954, opened 2016-04-11
Bugfix unit tests

olegabr:
Do not merge it yet! I'm using it to trigger travis build.


olegabr:
Ready to merge now.

spekary:
Unfortunately, the display for the test fired from the start page is now not working. It has reverted to the default output coming from the command line, which is hard to read. You need to use different configuration files for the travis test and the start page test I think.

olegabr:
@spekary, please, check this one

PR #969, opened 2016-05-10
bugfix->ReverseReferencesOfForwardExpansions

spekary:
There is a problem in how expansions are done currently, in that when you expand on a foreign key, and then with the object, look across the foreign key, and then look back, you see a null, which causes a Load to happen, and which loads a copy of the object looking backwards.

In other words, 

```
		$clauses = [QQ::Expand(QQN::Person()->ProjectAsManager)];
		$objPerson = Person::QuerySingle(QQ::All(), $clauses);
		$objProject = $objPerson->ProjectAsManager;
		$objPerson2 = $objProject->ManagerPerson;
```
$objPerson2 is not the same as $objPerson, plus an extra hidden query gets executed to fill in ManagerPerson when you ask for it. This can make some tables very slow to load, and make complex relationships difficult to manage.

This pull request fixes things so that $objPerson2 will be a reference to the same object as $objPerson, and makes it so that whenever you do a forward expansion, the matching reverse expansion is immediately and automatically populated with the forward looking object, so no additional query is generated.

You can still override this behavior by simply adding an expansion clause across the reverse reference. For example, adding:
```
$clauses = [QQ::Expand(QQN::Person()->ProjectAsManager->ManagerPerson)];
```
will cause $objPerson2 to again be a copy of $objPerson, instead of the same object.

### DIFF
--- a/includes/codegen/templates/db_orm/class_gen/instantiation_methods.tpl.php
+++ b/includes/codegen/templates/db_orm/class_gen/instantiation_methods.tpl.php
@@ -22,12 +22,22 @@
 		 * @param DatabaseRowBase $objDbRow
 		 * @param string $strAliasPrefix
 		 * @param QQNode $objExpandAsArrayNode
-		 * @param QBaseClass $arrPreviousItem
-		 * @param string[] $strColumnAliasArray
+		 * @param QBaseClass $objPreviousItemArray Used by expansion code to aid in expanding arrays
+		 * @param string[] $strColumnAliasArray Array of column aliases mapping names in the query to items in the object
 		 * @param boolean $blnCheckDuplicate Used by ExpandArray to indicate we should not create a new object if this is a duplicate of a previoius object
+		 * @param boolean $strParentExpansionKey If this is part of an expansion, indicates what the parent item is
+		 * @param boolean $strParentExpansionObject If this is part of an expansion, is the object corresponding to the key so that we can refer back to the parent object
 		 * @return mixed Either a <?= $objTable->ClassName ?>, or false to indicate the dbrow was used in an expansion, or null to indicate that this leaf is a duplicate.
 		*/
-		public static function InstantiateDbRow($objDbRow, $strAliasPrefix = null, $objExpandAsArrayNode = null, $objPreviousItemArray = null, $strColumnAliasArray = array(), $blnCheckDuplicate = false) {
+		public static function InstantiateDbRow(
+				$objDbRow,
+				$strAliasPrefix = null,
+				$objExpandAsArrayNode = null,
+				$objPreviousItemArray = null,
+				$strColumnAliasArray = array(),
+				$blnCheckDuplicate = false,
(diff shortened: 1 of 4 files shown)


includes/codegen/templates/db_orm/class_gen/instantiation_methods.tpl.php
```diff
@@ -22,12 +22,22 @@
 		 * @param DatabaseRowBase $objDbRow
 		 * @param string $strAliasPrefix
 		 * @param QQNode $objExpandAsArrayNode
-		 * @param QBaseClass $arrPreviousItem
-		 * @param string[] $strColumnAliasArray
+		 * @param QBaseClass $objPreviousItemArray Used by expansion code to aid in expanding arrays
+		 * @param string[] $strColumnAliasArray Array of column aliases mapping names in the query to items in the object
 		 * @param boolean $blnCheckDuplicate Used by ExpandArray to indicate we should not create a new object if this is a duplicate of a previoius object
+		 * @param boolean $strParentExpansionKey If this is part of an expansion, indicates what the parent item is
```
olegabr: type boolean and prefix str conflict

includes/codegen/templates/db_orm/class_gen/instantiation_methods.tpl.php
```diff
@@ -22,12 +22,22 @@
...
+		 * @param boolean $strParentExpansionObject If this is part of an expansion, is the object corresponding to the key so that we can refer back to the parent object
```
olegabr: same here, but alse, prefix should be obj may be?

spekary: Good catch. Fixed.

includes/codegen/templates/db_orm/class_gen/instantiation_methods.tpl.php
```diff
@@ -25,8 +25,8 @@
 		 * @param QBaseClass $objPreviousItemArray Used by expansion code to aid in expanding arrays
 		 * @param string[] $strColumnAliasArray Array of column aliases mapping names in the query to items in the object
 		 * @param boolean $blnCheckDuplicate Used by ExpandArray to indicate we should not create a new object if this is a duplicate of a previoius object
-		 * @param boolean $strParentExpansionKey If this is part of an expansion, indicates what the parent item is
-		 * @param boolean $strParentExpansionObject If this is part of an expansion, is the object corresponding to the key so that we can refer back to the parent object
+		 * @param string $strParentExpansionKey If this is part of an expansion, indicates what the parent item is
+		 * @param boolean $objExpansionParent If this is part of an expansion, is the object corresponding to the key so that we can refer back to the parent object
```
olegabr: what about type here? mixed?

spekary: yep

PR #972, opened 2016-05-10
Added Debug Mode to RenderComment

vaibhav-kaushal:
Added Debug Mode to ensure production environments do not give out more details than required. Also, BlockControls can now show Templates (helps debugging which Template is being rendered, in case multple can be used in the project)

### DIFF
--- a/includes/base_controls/QControlBase.class.php
+++ b/includes/base_controls/QControlBase.class.php
@@ -1635,7 +1635,16 @@
 		 * @return string
 		 */
 		public function RenderComment($strType) {
-			return  QHtml::Comment( $strType . ' ' . get_class($this) . ' ' . $this->strName . ' id:' . $this->strControlId);
+			// We render the comment only if the debug mode is on
+			if (defined('__DEBUG_MODE__') && __DEBUG_MODE__ == true) {
+				if($this instanceof QBlockControl) {
+					return  QHtml::Comment( $strType . ' ' . get_class($this) . ' Template: ' . $this->Template . $this->strName . ' id:' . $this->strControlId);
+				} else {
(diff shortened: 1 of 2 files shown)


includes/base_controls/QControlBase.class.php
```diff
@@ -1635,7 +1635,16 @@ public function RenderWithName($blnDisplayOutput = true) {
 		 * @return string
 		 */
 		public function RenderComment($strType) {
-			return  QHtml::Comment( $strType . ' ' . get_class($this) . ' ' . $this->strName . ' id:' . $this->strControlId);
+			// We render the comment only if the debug mode is on
+			if (defined('__DEBUG_MODE__') && __DEBUG_MODE__ == true) {
+				if($this instanceof QBlockControl) {
```
olegabr: it would be better to override the RenderComment method in the QBlockControl class, instead of using instanceof here. just better OOP style.

vaibhav-kaushal: Two reasons against it:

1. We have just one place where we are rendering the comment this way. 
2. We are using the instanceof to print the template path. 

This one is concise, in my opinion.

spekary:
Notice that the docblock for this function says "Render Comment if we are not in MINIMIZE mode". If you see QHTML::Comment, you will notice that nothing prints when we are in MINIMIZE mode. Production environments should be in MINIMIZE mode, and then comments won't print. This DEBUG_MODE check is not necessary.

vaibhav-kaushal:
I totally ignored it! It's like I have somehow developed ignorance towards the comments. Should we not introduce the `__MINIMIZE__` variable in the configuration.inc.php(.sample) file and set it to false?

spekary:
Good idea. Will post soon.

PR #974, opened 2016-05-10
bugfixAndFeatures-VirtualAttributes

spekary:
Adding and fixing the following:
- Aggregate clauses were not lowercasing the attribute name, and GetVirtualAttribute was not lowercasing before looking for an attribute. I added a central spot to convert a Virtual Attribute name into the name used in the query. Added unit tests that failed earlier and now succeed. You can now use any string as an attribute name, lower or upper case is fine.
- Added the ability to sort by virtual attributes, including virtual attributes created by Aggregate clauses. To do this, I changed how Aggregate clauses work so that they are registered as virtual attributes and use the new QQ::Func class. Also had to create a system to delay the evaluation of OrderBy clauses so that all the Virtual attribute expansions could register first. I also special cased virtual attributes so that Order By will use the alias instead of recreating the sql for the attribute. My research seems to suggest that this will work for all SQLs.
- QVitualAttributeColumn will now sort on the attribute by default.


olegabr:
It fails for all postgresql builds:
```
There was 1 error:
1) BasicOrmTests::testVirtualAttributeAliases
pg_query(): Query failed: ERROR:  column "t0.id" must appear in the GROUP BY clause or be used in an aggregate function
LINE 2:     "t0"."id" AS "a0",
            ^
/home/travis/build/qcubed/framework/includes/database/QPostgreSqlDatabase.class.php:243
/home/travis/build/qcubed/framework/includes/framework/QDatabaseBase.class.php:547
/home/travis/build/qcubed/framework/includes/framework/QModelTrait.trait.php:187
/home/travis/build/qcubed/framework/install/project/generated/model_base/ProjectGen.class.php:353
/home/travis/build/qcubed/framework/includes/tests/qcubed-unit/BasicOrmTest.php:406
/home/travis/build/qcubed/framework/includes/base_controls/QFormBase.class.php:553
FAILURES!
Tests: 112, Assertions: 840, Errors: 1.
```

spekary:
Going to repost this to fix some kind of strange merge conflict.

PR #977, opened 2016-05-11
Example rss_feed.php showed a parsing...

kukrik:
Logo address was invalid in RSS


olegabr:
not very good place for icon, but better then nothing :-)

PR #980, opened 2016-05-18
feature-QCubedJSRefactor

spekary:
Did some refactoring and simplification of the core QCubed.js code. Brought the server and ajax code into alignment as much as possible so they share much of the same code. Simplified the handling of checkable controls. Using built-in javascript and PHP functions to handle encoding of the ajax post data, rather than encoding it by hand. Added documentation.


spekary:
Just added a new mechanism to allow custom controls to insert post variables before posting. This mechanism is required for a fix to the CKEditor plugin to allow it to retrieve the text over Ajax.

PR #985, opened 2016-05-24
bugfix-characterEncoding

spekary:
Fixing a number of problems with character encoding that is not UTF-8. I think I have it working so that all variables are unpacked into the same encoding as QApplication::$EncodingType. Tricky stuff.

I would like some people to verify this before committing it, so consider this a work in progress.

In particular, please open the file vendor/qcubed/framework/assets/php/tests/ui-postvars.php in a browser and try it out. Someone please try it in IE. Also, try changing the encoding type at the top of the file to something like cyrillic and test again.

I am concerned in particular about the ability of JSON.stringify to encode non- utf-8 strings, and then have those strings get past through the server submit and ajax submit processes. Some help testing this would be appreciated.

### DIFF
--- a/assets/js/qcubed.js
+++ b/assets/js/qcubed.js
@@ -21,7 +21,7 @@ $j.fn.extend({
  * Queued Ajax requests.
  * A new Ajax request won't be started until the previous queued
  * request has finished.
- * @param object o Options.
+ * @param {object} o Options.
  */
 $j.ajaxQueue = function(o) {
     if (typeof $j.ajaxq === "undefined") {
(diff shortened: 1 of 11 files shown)


olegabr:
@spekary, I have no windows to test it. And there were no activity here for a long time. In this case we should merge this PR, and if it causes problems we would address it later.

spekary:
OK. The main problem this might cause is that we are now declaring the Content-Type with a "header" command at the end of a post. This will cause problems with some legacy code if someone was trying to declare a different content type through a Header command. There are some easy workarounds under the new code:
- Set QApplication::$ProcessOutput to false
- Set QApplciation::$CotnentType to the content type desired
- Send the header commands, send the output, and them simply call "exit" to prevent any QCubed processing.

All of these are used in the final commit to make sure the QCubed internal code works.

The nice thing this fixes is that before, when someone sets QApplication::$EncodingType, it didn't do anything on many servers because the server was sending out a default Content-Type header, which would cause the client to ignore the encoding declared in the HTML. A difficult thing to debug, and not always fixable depending on the hosting environment.

PR #1007, opened 2016-06-23
feature-AsyncAjax

spekary:
This adds the ability to send ajax events to the server asynchronously. By default, multiple ajax events are queued, and each one can not fire until the previous one returns a result and the result is processed. The primary reason for this is to allow the formstate to be refreshed.

However, there are times when you don't want to wait:
1. You are using a formstate handler that reuses the same formstate key for ajax events (the db handler does this now.) So, you know the formstate key is not going to change between ajax calls.
2. You are rapidly sending ajax events to give quick user feedback.

### DIFF
--- a/assets/js/qcubed.js
+++ b/assets/js/qcubed.js
@@ -22,9 +22,11 @@ $j.fn.extend({
  * A new Ajax request won't be started until the previous queued
  * request has finished.
  * @param {object} o Options.
+ * @param {boolean} blnAsync true to launch right away.
  */
-$j.ajaxQueue = function(o) {
-    if (typeof $j.ajaxq === "undefined") {
+$j.ajaxQueue = function(o, blnAsync) {
+    if (typeof $j.ajaxq === "undefined" || blnAsync) {
+        if (o.fnInit) o.fnInit(o);
         $j.ajax(o);
     } else {
         // see http://code.google.com/p/jquery-ajaxq/ for details
@@ -361,10 +363,11 @@ qcubed = {
      * @param {string} strEvent
      * @param {null|string|Object|Array} mixParameter
      * @param {string} strWaitIconControlId The id of the control's spinner.
+     * @param {boolean} blnAsync Whether to queue the ajax requests and processes serially (default), or do them async
      * @return {void}
      * @todo There is an eval() in here. We need to find a way around that.
      */
-    postAjax: function(strForm, strControl, strEvent, mixParameter, strWaitIconControlId) {
+    postAjax: function(strForm, strControl, strEvent, mixParameter, strWaitIconControlId, blnAsync) {
         var objForm = $j('#' + strForm),
             strFormAction = objForm.attr("action"),
             qFormParams = {};
@@ -448,7 +451,7 @@ qcubed = {
                     qcubed.processDeferredAjaxResponse(json);
                 }
             }
-        });
+        }, blnAsync);
 
     },
 
--- a/assets/js/qcubed.js
+++ b/assets/js/qcubed.js
@@ -363,7 +363,8 @@ qcubed = {
      * @param {string} strEvent
      * @param {null|string|Object|Array} mixParameter
      * @param {string} strWaitIconControlId The id of the control's spinner.
-     * @param {boolean} blnAsync Whether to queue the ajax requests and processes serially (default), or do them async
+     * @param {boolean} blnAsync Whether to queue the ajax requests and processes serially (default), or do them async.
+     *                  See QAjaxAction comments for more info
      * @return {void}
      * @todo There is an eval() in here. We need to find a way around that.
      */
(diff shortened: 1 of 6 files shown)


matthiaz:
@spekary You mention "each one can not fire until the previous one returns a result"
So doesn't that mean this is a Synchronous Ajax? If it was Assynchronous I wouldn't expect the results to come back in the same order. But rather, be fired in parallel

spekary:
@matthiaz, I was referring to how qcubed.js currently works by default. Yes, it makes synchronous Ajax calls. This pull request allows you to specify that you want particular events to come through asynchronously, bypassing the qcubed.js Ajax queue. They will still fire in order, but will not wait for one call to return before firing the next one.

The danger with this approach is that if an ajax call changes the key used to specify the formstate, weird things can happen. Essentially, some of the ajax calls will use an old formstate, which could possibly also change the formstate in unexpected ways. 

So, the programmer who wants ajax calls to be asynchronous must know that either the formstate KEY will not change (though the contents of the formstate could change), or that the contents of the formstate will not change in any meaningful way.

matthiaz:
Thanks for the clarification.

olegabr:
It was me who introduced the ajax queueing for qcubed in the first place. And it was because of very hard to figure out errors I've faced with it. The obvious one, when user clicks button very fast several times and handler changes some formstate - results are undefined. state would be rewritten several times in an unpredictable order. So, @spekary, why you open this door to the hell again? ;-)

spekary:
Heh, heh. The queue is a good thing. However, when you added the queue, I think the problem you ended up fixing was one that was fixable in a different way.

When you implemented the queue, there was a contributing problem in that the key for the formstate was changed on every ajax call. This, I assume was done long ago by the qcodo creator to simulate the FormState handler that put the entire form state into the form as a hidden control. That handler should rarely be used, if at all. Its slow and insecure, and has this problem where the javascript ajax queue is required. Because the formstate key was changing, and each ajax call was sending in its last known formstate key, then multiple ajax calls would use the same, outdated key, and not see the formstate changes caused by a previous event.

About a year ago, I changed the formstate handlers to re-use the same key. This is what makes it possible to bypass the queue in general. Even if the formstate changes, the next ajax event will get the new formstate because it will get it from the same database record, or location in the session, which was updated by the previous event. Since the key didn't change, when the next event sends in the same key, it will contain the updated formstate.

As I see it, the only problem that would occur without the queue is if the ajax events were causing form objects to be redrawn in such a way that the next event would be dependent on these new form objects. I can't really think of a scenario where that would happen, since the event was sent in before the user saw the changes, and so was based on the previous view anyway. The only other problem would be that the generic formstate handler requires the queue to function correctly.

The reason I am adding this now is that I am making a control that is rapid firing many events. The ajax queue is slowing down its respose time, because each event has to wait to complete and process its results before the next event can be sent to the server.

Note that this change does not remove the event queue. The event queue is still on by default. This change only allows bypassing the queue when the programmer specifically requests it by passing a final "true" value to a QAjaxAction or QAjaxControlAction.

assets/js/qcubed.js
```diff
@@ -361,10 +363,11 @@ qcubed = {
      * @param {string} strEvent
      * @param {null|string|Object|Array} mixParameter
      * @param {string} strWaitIconControlId The id of the control's spinner.
+     * @param {boolean} blnAsync Whether to queue the ajax requests and processes serially (default), or do them async
```
olegabr: Please, place here a warning that it should be used with care, and why.

includes/base_controls/_actions.inc.php
```diff
@@ -297,12 +299,13 @@ class QAjaxAction extends QAction {
 		 *                                                      client-side when the action occurs!
 		 */
 		public function __construct($strMethodName = null, $objWaitIconControl = 'default',
-		                            $mixCausesValidationOverride = null, $strJsReturnParam = "") {
+		                            $mixCausesValidationOverride = null, $strJsReturnParam = "", $blnAsync = false) {
```
olegabr: Please, document the new parameter.

includes/base_controls/_actions.inc.php
```diff
@@ -425,8 +429,8 @@ class QAjaxControlAction extends QAjaxAction {
 		 * @param string   $strJsReturnParam            Override for ActionParameter
 		 */
 		public function __construct(QControl $objControl, $strMethodName, $objWaitIconControl = 'default',
-		                            $mixCausesValidationOverride = null, $strJsReturnParam = "") {
-			parent::__construct($objControl->ControlId . ':' . $strMethodName, $objWaitIconControl, $mixCausesValidationOverride, $strJsReturnParam);
+		                            $mixCausesValidationOverride = null, $strJsReturnParam = "", $blnAsync = false) {
```
olegabr: Please, document the new parameter.

olegabr:
I understand your reasons, thank you for explanation. Just document these new parameters please with a proper statement about possible danger of it's usage.

spekary:
Did the documentation. Also, fixed up QFileFormStateHandler to be async compatible. So, the only danger here is using the standard QFormStateHandler, which most of the time people shouldn't be using.

PR #1018, opened 2016-06-28
fixes #458 (Security attributes for many qcubed files are incorrect)

### LINKED ISSUE #458: Security attributes for many qcubed files are incorrect
In the already merged commit
https://github.com/vaibhav-kaushal/qc-framework/commit/8c35573ea15e923801f3766f31f15df9b40f3cc0 there are a lot of sentences like that: ` .gitignore 100644 → 100755`

It should be fixed back to 644 for all files listed in this commit.

olegabr:


spekary:
Looks like travis is having permission problems.

olegabr:
fixed now

spekary:
Can you summarize why you are preventing access from general use? I am concerned that there are situation where someone may be using composer to get files as a user that is not the same user as the programmer. For example, if you run composer as super-user because of some composer problem, then you will be forced to chmod everything before you can use it as non-super user. Also, what about codegen? Since codegen runs from the browser, the browser process will be the owner of the files generated. 

And as I think about it, won't preventing other users from reading the files also prevent the php or hhvm process from reading it? Someone would have to be sure the programmer and the process were in the same group, correct? 

I haven't tested this, so just thinking off the top of my head. I assume you are trying to improve security. Wouldn't it be better to make this part of a deployment script?

olegabr:
You are right - It better be additionally ensured in the deployment script. In this PR I just set standard permissions for files exposed for a web server 644. That means that everyone can read it, but only the owner can modify. And no one can execute. Also, I've set 755 permissions on directories, that means that it can be executed (traversed) and read by everyone, but new files can be created there only by  owner.

We already have some checks in the `composer.json` to ensure some folders to be writable for development. Look for the `writePermission` there. And some check are in a php page that checks the installation.

PR #1024, opened 2016-06-29
Added JSON and JSONB support for PostgreSQL Database

vaibhav-kaushal:
Only basic support so far. But enough to not cause errors or disrupt functionality for Codegen and existing code.


includes/framework/QType.class.php
```diff
@@ -68,13 +68,23 @@ public function __construct($strMessage, $intOffset = 2) {
...
+		/** Float Type */
 		const Float = 'double';
+		/** Boolean Type */
 		const Boolean = 'boolean';
+		/** Object Type */
 		const Object = 'object';
+		/** Array Type */
 		const ArrayType = 'array';
+		/** JSON - String Type but valid JSON Text */
+		const Json = 'json';
```
spekary: The problem with this is that all the other types are standard PHP types. As we move towards PHP7, we will be using these types to annotate the code. Having a type that is not a standard PHP type could cause a significant problem.

vaibhav-kaushal: If we are going to support database-specific types, then we need to declare them so in the DatabaseBase class. Do you have any suggestion on how we can achieve that? 

If you think we should drop it for now, I'm okay with that so far. However, the modifications in QPostgreSqlDatabase class are required to not let Code Generation fail. We can change that behavior to 'text' treatment for JSON types. But that would be a little troublesome later in code generation. How about we get a `QDbDataType` for things like that?

includes/framework/QDatabaseBase.class.php
```diff
@@ -1299,5 +1299,7 @@ public function __get($strName) {
 		const Float = "Float";
 		/** Boolean */
 		const Bit = "Bit";
+		/** JSON (for Postgresql currently) */
+		const Json = "JSON";
```
spekary: Again, I think we need a new architecture here to support data types that are sepecific to a database implementation. Is it possible to subclass here inside the Postgresql adapter and add these custom types there? It generally bad practice to make parent class aware of subclasses.

includes/framework/QType.class.php
olegabr: @vaibhav-kaushal why we need it in QType at the first place? where it is used?

vaibhav-kaushal: I was thinking that we could use type casting for conversion. But I have pulled it back. It does not make much sense.

includes/framework/QDatabaseBase.class.php
vaibhav-kaushal: I believe, JSON is coming into more and more databases anyway. However, I think that it could do better in another class. 
Let me see about that.

vaibhav-kaushal:
@spekary and @olegabr Can you guys help fix this problem with Travis-CI?

olegabr:
It looks like `QPostgreSqlDatabaseFieldType` class is not accessible from the `QCodeGenBase` file. Make sure to include it in some way. May be you should add it to the auto-include file?

olegabr:
It is broken only for mysql builds. pg builds are ok:
```
PHP Fatal error:  Class 'QPostgreSqlDatabaseFieldType' not found in /home/travis/build/qcubed/framework/includes/codegen/QDatabaseCodeGen.class.php on line 1334
```
It looks like there is some conditional that includes the file with the class `QPostgreSqlDatabaseFieldType` only for postgresql build. try to add the class `QPostgreSqlDatabaseFieldType` to the auto-inclusion file, like it did for the `QPostgreSqlDatabase` class.

vaibhav-kaushal:
@olegabr If you look here: https://github.com/qcubed/framework/pull/1024/commits/f3db4517db7dce57ee86749be3ce75cb85d588da#diff-d1db6ec4153570a42d5a8428613a753bR39

I have already tried that and it failed.

vaibhav-kaushal:
@spekary I might need your help here in setting up the Travis CI automated tests.

includes/db_specific/PostgreSql/QPgQuery.class.php
```diff
@@ -0,0 +1,224 @@
...
+	class QPgQConditionJsonbEqual extends QQConditionComparison {
+		/**
+		 * @param QQColumnNode $objQueryNode
+		 * @param string       $strValue
+		 *
+		 * @throws Exception
+		 * @throws QCallerException
+		 * @throws QInvalidCastException
+		 */
+		public function __construct(QQColumnNode $objQueryNode, $strValue) {
```
spekary: I am working on some changes, but just wanted to comment on this. This appears unnecessary here. Just use the standard QQ::Equals for this. Or am I missing something? Also, you don't need something special for jsonb vs json. The Postgres documentation says the two columns will produce the same results for queries, its just a matter of the internal storage format. JsonB allows postgres to store the json already parsed, whereas json type will force postgres to parse the json every time you do an operation. The results of the operation are identical, just not the performance.

spekary:
See the pull request I sent to the branch in your repo

Issue #875, opened 2016-01-09
Examples should use QDataGrid2


spekary:
QDataGrid is obsolete. We should migrate all examples to QDataGrid2.

kukrik:
Yes!

spekary:
Working on this one.

PR #1026, opened 2016-07-05
bugfix_DataGrid2InExamples

spekary:
Extensive work moving all the example code to use QDataGrid2.

Spurred some additional changes to QSimpleTable, QDataGrid2 and some other classes to make them work with the examples.

Looks like this is not currently mergeable, so I need to look at that. Don't merge yet.


spekary:
OK, It should be able to check in now.

spekary:
Fixes #875

Issue #896, opened 2016-01-22
QDataGrid2 connects to  external site


nlomb:
just found out that QDataGrid2 connects to an external site, see code
file: QDataGrid2Base.class.php
```php
...
	public function __construct($objParentObject, $strControlId = null)	{
		try {
			parent::__construct($objParentObject, $strControlId);

			$this->AddCssFile('//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css');

			$this->AddActions();

		} catch (QCallerException  $objExc) {
			$objExc->IncrementOffset();
			throw $objExc;
		}
	}
...
```
I think it should be possible to deactivate this, in a closed environment or with no internet connection this may produce errors or strange behaviour.

spekary:
Good point. We should download this then and include it in our distribution. FontAwesome is MIT license, so shouldn't be a problem.

PR #1028, opened 2016-07-05
bugfix-QDataGrid2Font

spekary:
Making FontAwesome installable as an option. FontAwesome is now open-source licensed for general use.


spekary:
Fixes #896

Issue #1025, opened 2016-07-04
QOnEvent parameter order inconsistent with other events


spekary:
QOnEvent uses

```
$strEventName, $strSelector = null, $strCondition = null, $intDelay = 0
```

All other events use:

```
$intDelay = 0, $strCondition = null,  $strSelector = null, 
```

This made some sense when QOnEvent was first created to specifically handle delegated events, but now all events can handle delegated events, and QOnEvent is more useful to handling custom event triggers and event types that we have not created a QEvent for.

spekary:
Fixed #1029

PR #1029, opened 2016-07-08
bugfix-QOnEventParamOrder

spekary:
Fixing parameter order of QOnEvent, and some language around the usage of it.

### DIFF
--- a/includes/base_controls/QControlBase.class.php
+++ b/includes/base_controls/QControlBase.class.php
@@ -45,8 +45,8 @@
 	 * when something in the database changes.
 	 *
 	 * QControls are the base objects for actions to be attached to events. When attaching actions to multiple objects
-	 * of the same type, considering using the QOnEvent trigger for you action, as it can be more efficient in certain
-	 * cases.
+	 * of the same type, considering attaching the event to a a parent object and using event delegation for your action,
+	 * as it can be more efficient in certain cases.
 	 *
 	 * QControls can trigger validation and are part of the validation system. QControls that are not Enabled or not
 	 * Visible will not go through the form's Validation routine.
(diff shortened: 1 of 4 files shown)


spekary:
Fixes #1025

includes/base_controls/QControlBase.class.php
```diff
@@ -45,8 +45,8 @@
 	 * when something in the database changes.
 	 *
 	 * QControls are the base objects for actions to be attached to events. When attaching actions to multiple objects
-	 * of the same type, considering using the QOnEvent trigger for you action, as it can be more efficient in certain
-	 * cases.
+	 * of the same type, considering attaching the event to a a parent object and using event delegation for your action,
```
olegabr: 'a' duplication here

includes/base_controls/_events.inc.php
```diff
@@ -328,13 +328,13 @@ class QOnEvent extends QEvent{
...
 		 * @param int  $strEventName
-		 * @param string $strSelector
 		 * @param string $strCondition
 		 * @param int  $intDelay
+		 * @param string $strSelector
 		 *
 		 * @throws Exception|QCallerException
 		 */
-		public function __construct($strEventName, $strSelector = null, $strCondition = null, $intDelay = 0) {
+		public function __construct($strEventName, $strCondition = null, $intDelay = 0, $strSelector = null) {
```
olegabr: As you have mentioned in #1025, the right order in other events is:
```
$intDelay = 0, $strCondition = null,  $strSelector = null
```

But here you have placed `$strCondition` to the front. Why?

spekary: Because I am too tired to do this. Thanks for catching this. Will fix it in a bit.

spekary:
Should be ready to merge now.

PR #1033, opened 2016-07-22
Consistent CACHE and UPLOADS dir naming fixes #889

### LINKED ISSUE #889: Cache and uploads folders in assets needs to be rethinked
Moved from #870 

As @kukrik mentioned already, we need two cache folders: one for php-accessible things, and it should be full filesystem path with `__DOCROOT__`, and another one for web-accesible goods, and it should be relative to site domen, e.g. for `http://example.com/project/assets/cache/myphoto.png` it should be `/project/assets/cache`

For the first one we have `__TMP__`. For the second we don't have one place, but only images cache in the `__APP_IMAGE_CACHE__` define. We should have `__APP_CACHE_ASSETS__ `instead and, may be, `__APP_IMAGE_CACHE_ASSETS__` defined as `__APP_CACHE_ASSETS__  . '/images'`. And we need to have `__APP_CACHE__` as a full-path version of the  `__APP_CACHE_ASSETS__` with a `__DOCROOT__` included. The `__APP_IMAGE_CACHE__` becomes a full-path version of the `__APP_IMAGE_CACHE_ASSETS__` define.

Uploads are different. All other caches are volatile and can be cleared in any time. Uploads are not. They are a part of user expirience and should be preserved. That is a reason to maintain it separately from cache directory. It means that the current `__APP_UPLOAD__` is done right. But it lacks it's `ASSETS` counterpart: `__APP_UPLOAD_ASSETS__` that is needed to be used by client code to form URI-s of uploaded items. The original `__APP_UPLOAD__` is used by the code that needs to work with file in a filesystem.

olegabr:


travis/configuration.inc.php
```diff
@@ -7,11 +7,16 @@
 	// Feel free to use, change or ignore.
 	define('SERVER_INSTANCE', 'dev');
 
+	/** Define constants here to control the differences between your development, test and production environments. */
+
 	switch (SERVER_INSTANCE) {
 		case 'dev':
+			define('__DESIGN_MODE__', 1);
```
spekary: Why are we doing this for the travis build?

includes/framework/QInstallationValidator.class.php
```diff
@@ -161,6 +161,34 @@ public static function Validate() {
 			$result[] = $obj;
 		}
 
+		if (!file_exists(__IMAGE_CACHE__)) {
```
spekary: Should this be __IMAGE_CACHE_ASSETS__ ??

olegabr: No, `_assets` version is to construct URLs. here we need full path in a filesystem with `__DOCROOT__` applied.

travis/configuration.inc.php
olegabr: For consistency with `configuration.inc.php` in the project. These two files should be as close as possible to ensure proper testing.

PR #1037, opened 2016-07-27
QSubSql generates invalid sql in some cases.

matthiaz:
You only need to wrap a subsql statement with braces (  ) when it is an actual query. 
For example: 
`SELECT * FROM table1 WHERE col1 = (SELECT count(id) FROM table2)`

If you want to use QSubSql for something other that pure queries, you can run into problems.
For example: 
```
Client::QuerySingle(
          QQ::All(),
          QQ::Clause(
              QQ::Count(QQ::SubSql('DISTINCT t0.client_id'), "count_clients")
          )
    );
```
Will generate
`SELECT COUNT( (DISTINCT client_id) ) as distinct_client_count FROM client;`

but that is invalid sql, it should be
`SELECT COUNT( DISTINCT client_id ) as distinct_client_count FROM client;`
Without the braces ( )

### DIFF
--- a/includes/qcubed/_core/framework/QQuery.class.php
+++ b/includes/qcubed/_core/framework/QQuery.class.php
@@ -1326,7 +1326,11 @@
 				if (!is_null($this->objParentQueryNodes[$intIndex]))
 					$strSql = str_replace('{' . $intIndex . '}', $this->objParentQueryNodes[$intIndex]->GetColumnAlias($objBuilder), $strSql);
 			}
-			return '(' . $strSql . ')';
+			if(stripos($strSql, 'SELECT')===0) {
+				return '(' . $strSql . ')';
+			} else {
+				return $strSql;
+			}
 		}
 	}
 


spekary:
Is it possible to put this in unit test? Will the invalid SQL generate a SQL exception that we can catch?

matthiaz:
I can give it a shot if someone points me in the right direction (I have no idea where to make these unit tests)

the sql exception that gets thrown is
```
MariaDB> select COUNT((DISTINCT client.client_id)) FROM client;

ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'DISTINCT client.client_id)) FROM client' at line 1
```

So yeah, qcubed will throw an error when running:

```
$objDbResult = $objQueryBuilder->Database->Query($strQuery);
```

It depends on the database driver which exception will be thrown. In my case, this is a **QMySqliDatabaseException**

spekary:
framework/includes/tests/qcubed-unit/QDatabaseTest.php. Add a test case function at the bottom. It will get tested under MySql and Postgresql by the travis build.

Issue #1048, opened 2016-08-03
error with PHP 7


nlomb:
just upgraded my demo - server to ubuntu 16.04 LTS, this comes with php 7
got the following error:
```
Error in PHP Script /resources/vendor/qcubed/framework/assets/php/_devtools/start_page.php
Uncaught TypeError: Argument 1 passed to QcubedHandleException() must be an instance of Exception, instance of Error given in /usr/data/share/www/.develop/qcubed-demo/resources/vendor/qcubed/framework/includes/error.inc.php:15
Stack trace:
#0 [internal function]: QcubedHandleException(Object(Error))
#1 {main}
  thrown

Error Type: E_ERROR

Source File: /usr/data/share/www/.develop/qcubed-demo/resources/vendor/qcubed/framework/includes/error.inc.php Line: 15

Line 10:    			$this->MultiLine = $blnMultiLine;
Line 11:    		}
Line 12:    	}
Line 13:    
Line 14:    	// Exception and Error Handling
Line 15:    	function QcubedHandleException(Exception $__exc_objException) {
Line 16:    		if (class_exists('QApplicationBase'))
Line 17:    			QApplicationBase::$ErrorFlag = true;
Line 18:    
Line 19:    		global $__exc_strType;
Line 20:    		if (isset($__exc_strType))

Call Stack:

#0 /usr/data/share/www/.develop/qcubed-demo/resources/vendor/qcubed/framework/includes/error.inc.php(194): QcubedHandleError()
#1 (): QCubedShutdown()

Variable Dump: Show/Hide

Error Report Generated: Wednesday, August 3 2016, 1:14:42 AM
PHP Version: 7.0.8-0ubuntu0.16.04.2; Zend Engine Version: 3.0.0; QCubed Version: 3.0.0 Beta Release (QCubed 3.0.0)
Application: Apache/2.4.18 (Ubuntu); Server Name: server-03
HTTP User Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:47.0) Gecko/20100101 Firefox/47.0 
```

vmakre:
Line 15:        function QcubedHandleException(Exception $__exc_objException) {

change to 
	            function QcubedHandleException(Throwable $__exc_objException) {

i really don't understand this line , but works for me.

nlomb:
@vmakre 
had also to do
```bash
sudo apt-get install php-mbstring php7.0-mbstring php-gettext libapache2-mod-php7.0
```
[...]

PR #1057, opened 2016-08-09
Fixing the PHP7 Error for error.inc.php

vaibhav-kaushal:
Fixing the PHP7 Error for error.inc.php

Refers #1048


spekary:
We need to test how it reacts when given a fatal error, but since this change doesn't break anything I checked it in.

vaibhav-kaushal:
@spekary Thanks

PR #1059, opened 2016-08-10
Renamed some readme files to README.md and added content

vaibhav-kaushal:
Renamed some readme files to README.md and added content


olegabr:
So much work! Thank you, @vaibhav-kaushal.

PR #1068, opened 2016-08-16
Update object_delete.tpl.php

kpirbhai:
Fix QWatcher::MarkTableModified call so that the database name isn't hard coded. This will stop problems when development and production databases have different names.

### DIFF
--- a/includes/codegen/templates/db_orm/class_gen/object_delete.tpl.php
+++ b/includes/codegen/templates/db_orm/class_gen/object_delete.tpl.php
@@ -55,7 +55,7 @@
 
 			$this->DeleteFromCache();
 			if (static::$blnWatchChanges) {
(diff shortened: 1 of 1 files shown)


includes/codegen/templates/db_orm/class_gen/object_delete.tpl.php
```diff
@@ -55,7 +55,7 @@ public function Delete() {
 
 			$this->DeleteFromCache();
 			if (static::$blnWatchChanges) {
-				QWatcher::MarkTableModified ('<?= QApplication::$Database[$objTable->OwnerDbIndex]->Database ?>', '<?= $objTable->Name ?>');
+				QWatcher::MarkTableModified($this->GetDatabaseName(), '<?= $objTable->Name ?>');
```
spekary: Try this instead:
```
				QWatcher::MarkTableModified (static::GetDatabase()->Database, '<?= $objTable->Name ?>');
```

PR #1069, opened 2016-08-16
Update object_save.tpl.php

kpirbhai:
Changed the way QWatcher::MarkTableModified is a called so that the database name isn't hard coded. This change will stop problems from occurring when the development environments and production environments have different database names.


spekary:
Please combine your changes into one pull request.

Also, there is no GetDatabaseName function. Do this instead in the templates:
```
QWatcher::MarkTableModified (static::GetDatabase()->Database, '<?= $objTable->Name ?>');
```
That should do what you want.

PR #1070, opened 2016-08-16
fixed codegen template db name hardcoded

kpirbhai:
Changed the codegen template files so that the database name isn't hard coded. This will fix problems where the database names are different between the development and production environments.
* My first pull request ever!!!!  :)


olegabr:
@kpirbhai Thank you!

spekary:
@kpirbhai Yes, a great addition, good job!

Issue #1063, opened 2016-08-12
Namespace Discussion


spekary:
This is a discussion on what the namespaces and the directory structure should look like in the next version of QCubed.
# Background

PSR4 requires that namespaces and the folder structure be the same. The reason for this requirement is primarily so that composer's auto-generated autoloader knows where to look for a class. It just looks at its namespace and use the namespace to create a path. Secondary to this is supportability, in that it allows humans to know where the file lives as well, but that is not such a big deal with modern IDEs that can easily find a file.

QCubed started before all the PSRs, and before even composer, and has always used a class map that lives in one of the qcubed.inc.php files. The desire to make QCubed more modular, allowing pieces of qcubed to be used without requiring the whole framework, may ultimately drive us to doing what PSR4 wants us to do. Being a full-stack framework, QCubed does not lend itself well to modularity, but we can try. Note that PSR4 has some flexibility, in that you can specify a base directory, meaning that the first portion of the namespace path does not have to directly map to a file structure. We get to define to some extent what portion that is.
# Strategy

As much as possible, logically self-contained pieces that have few dependencies should be in their own namespace. Whatever dependencies do exist, the composer.json file can refer to those. We need to balance the desire for modularity with the desire to keep things as simple as possible.
# Major Functional Areas

Architecturally, major functional areas of QCubed include:
- **Base classes**, that everything depends on. These live in framework, with other things:
  `QBaseClass`, `QDateTime`, `QException`, `QType`
- **ORM**, which has the following sub-parts and dependencies
  - Code generator, depending on:
    - `QCodeGen`, `QCodeGenBase`, and `QDatabaseCodeGen`
    - Support classes, like `QSqlColumn`, `QSqlTable`, `QReference`, `QReverseReference`, etc. Note that these classes are shared with the ui generation code.
    - code generation templates 
  - Database adapters and the database adapter base class
  - `QQuery`, and all the classes that live inside of there. Also the QQuery database specific extensions that now live in _db_ext_.
  - QModel.trait.php, which has some code shared between the generated base classes
  - The generated model classes, including
    - generated model base classes
      - dependency on some support classes, like QCache and QCacheProvider, but unknown as to what value those provide currently because of recent performance improvements.
    - generated node base (QQN class)
    - generated autoloader class map (_class_paths.inc.php)
    - The generated model subclass stubs that are one-time generated and avaialble for user editing, these live in the project/includes/model directory currently
  - Settings in the configuration.inc.php file
  - Unit tests
[...]

PR #1071, opened 2016-08-17
[WIP] QFormBase - add container

lackovic10:
This is an attempt to add a container to the `QFormBase` class, which accessible inside the `QForm` methods, as well as in `QPanels` and `QControls`.
Wanted to type hint the container object to the [Psr\Container\ContainerInterface](https://github.com/php-fig/container), but the standard is not approved yet

### DIFF
--- a/includes/base_controls/QFormBase.class.php
+++ b/includes/base_controls/QFormBase.class.php
@@ -19,6 +19,10 @@
 		///////////////////////////
 		// Static Members
 		///////////////////////////
+
+        /** @var Object $container */
+        protected static $container;
+
 		/** @var bool True when css scripts get rendered on page. Lets user call RenderStyles in header. */
 		protected static $blnStylesRendered = false;
 
@@ -202,6 +206,14 @@
 		 */
 		protected $intNextControlId = 1;
 
+		/**
+		 * @return Object
+		 */
+        public function getContainer()
+        {
+            return self::$container;
+        }
+
 		/////////////////////////
 		// Helpers for AjaxActionId Generation
 		/////////////////////////
@@ -368,7 +380,9 @@
 		 * @throws QInvalidFormStateException
 		 * @throws Exception
 		 */
-		public static function Run($strFormClass, $strAlternateHtmlFile = null, $strFormId = null) {
+		public static function Run($container, $strFormClass, $strAlternateHtmlFile = null, $strFormId = null) {
+            self::$container = $container;
+
 			// See if we can get a Form Class out of PostData
 			$objClass = null;
 			if ($strFormId === null) {
@@ -386,7 +400,7 @@
 					self::InvalidFormState();
 				}
 			}
-			
+
 			if ($objClass) {
 				// Globalize
 				global $_FORM;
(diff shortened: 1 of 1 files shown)


lackovic10:
@spekary what do you think about this solution ?
this will break all the `QForm::Run()` calls, i can add it as an optional parameter but it's not a good practice.
also, should some container implementation be included?

lackovic10:
@olegabr @vaibhav-kaushal check it out as well
why can't i assign people to the PR ?

spekary:
We are in late beta on the current release, so major breaking features are not going in the beta-3. Minor features still have some room if they have little risk.

I think the best thing to do here is to have you develop on your own branch in your own respository, and make frequent posts as you develop. As this matures, we can make this an experimental branch, and then eventually a 3.1 or 4.0 branch depending on how hard it is to port.

But nothing can go in to the qcubed repo that does not pass the travis unit tests at a minimum. You will need to possibly change the tests to have them work. Also, once this gets out of experimental, the examples will also need to be modified so that people can learn what to do.

Can you put the container at the end of the list, and default it to null? If we can make this optional, that will help with backward compatibility.

Also, I need to see examples of the container fleshed out. My biggest worry is whether the content is going to need to be serialized with the form state. If the container is just a container of function calls, then this is fine. However, if the container is going to itself contain data that is modified by the program, and that data needs to be recalled later, then this is all going to have to go into the formstate, which means the container needs to be saved as a variable inside the form object, and always referred to through the form object. The container cannot save lambdas, or callbacks. It CAN save callables, but you will need to hook into the Sleep and Wakeup calls to serialize those specially. If callback saving is needed, you will need to do a trick where you serialize a code or string that represents the callback, and restore the pointer to the callback when the form wakes up. Usually this involves creating some kind of static callback table which gets initialized whenever the programs starts. GoLang uses this trick.

In some of your previous examples, you always asked the form for the container. As long as the container comes from the saved container, that will work.

lackovic10:
when do you plan to release 3.0 ? maybe i can add the container i'm using but i need to ask my boss. it's very simple, one adds the services manually inside the class but works well for us so far.
it can be added as an optional parameter of course, to start with.
i'm not a fan of optional parameters in general so i think in later versions it should maybe be a non optional one.
the idea of adding my ideas to an experimental branch is something i also like.
i'll try to find some time to add the code generator as a separate component.
have you considered naming the repository `qcubed/qcubed` instead of `qcubed/framework` ?
the fork repository is `lackovic10/framework` which seems a little odd to me

spekary:
We are already past our planned release. The main issue I think is creating a new landing page in github and putting all the support info there. Our bug finding rate is going down rapidly, so I think we are pretty stable at this point.

Instead of an optional RUN parameter, you could create a global before you call Run, but its essentially the same thing.

You can do all of this in your private version of QForm by putting your code in the /project/includes/controls/QForm.class.php file. There you can setup your own Run function, and add your own member variables to QForm, and then call the parent Run function. You can experiment with your ideas there without impacting the framework at all. That is part of the beauty of the scaffolding. Almost everything is customizable through subclassing, inheritance and overriding.

spekary:
Regarding qcubed/qcubed, see #1063. That is a proposal there.

PR #1075, opened 2016-08-19
clean up file endings

kpirbhai:
Cleaned up the endings of all files to remove extraneous ?>
Also removed the ending ?> from the codegen templates


olegabr:
@kpirbhai thank you for this work! How have you did it? With some script or command? Can you please share it? May be we can include it into our testing workflow in order to force this style for all new additions.

vaibhav-kaushal:
@kpirbhai If you did it via script, we would love to see that one :) And this was pending since long!

kpirbhai:
Lol. No script. I just literally went through every file. It was bugging me.  ;)

vaibhav-kaushal:
@kpirbhai That must have taken some patience! Thanks :)

nlomb:
@olegabr 
script to remove extraneous "?>"  or to refactor some other things ?
would bash/sed/awk be appropriate ?
can do if needed.

olegabr:
@nlomb  script to remove extraneous "?>"  would be enough to start. then we can add one more check for travis: execute this script against the whole codebase and then call something like `diff -rq qcubed/ qcubed.processed/` and treat any difference as an error. It would prevent any regression in a sense of extraneous "?>"

PR #1077, opened 2016-08-25
bugfix-QDbBackedSessionHandler

spekary:
The base64 encoder was at sometimes producing a null result, which was causing the session to be dropped. This new version relies on a binary column instead of a text column and encrypts and compresses the data instead. This results in much more compact data with no dropped sessions and better security.

### DIFF
--- a/includes/framework/QDbBackedSessionHandler.class.php
+++ b/includes/framework/QDbBackedSessionHandler.class.php
@@ -30,6 +30,12 @@
 		 */
 		protected static $strSessionName = '';
 
+		/** @var bool Whether to encrypt the session data. Highly recommended whenever sessions can authenticate a user. */
+		public static $blnEncrypt = true;
+
+		/** @var bool Whether to compress the session data. */
+		public static $blnCompress = true;
+
 
 		/**
 		 * @static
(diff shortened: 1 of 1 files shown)


olegabr:
@spekary, this PR bundles session control with cryptography and compression. But it is not always required and often is CPU-intensive. So, there are should be a possibility to turn it off, or, better, replace with something different. So, at minimum, I want to see here some configurable in the `configuration.inc.php` options to turn cryptography and/or compression off.

nlomb:
just updated my dev environment, 
had to change database column "qc:session.data" from "(medium)text" to (medium)"blob".
should be mentioned in "configuration.inc.sample.php"

olegabr:
@nlomb why this change is required? which kind of error you have seen?

nlomb:
@olegabr 
see: https://github.com/qcubed/framework/pull/1077/commits/acb6397485d504b6a057e4e97cf2071f0e679f31
"This new version relies on a binary column instead of a text column and encrypts and compresses the data instead."

Error message is:
```
PHP Fatal error:  Uncaught QMySqliDatabaseException: MySqli Error: Incorrect string value: '\\x9C\\x03\\x00\\x00\\x00\\x00...' for column 'data' at row 1 in <MYDOCROOT>/resources/vendor/qcubed/framework/includes/database/QMySqliDatabase.class.php:68
```
in "/var/log/apache2/error.log"
 changing this column to "blob" fixes this (in LAMP environment)

Also the configuration variables in https://github.com/qcubed/framework/pull/1077/commits/f000a1ff5bd64f1b3ec34c09fe259a4bcc5a9c3f lines 33 - 38 should be moved to "configuration.inc.sample.php"

vaibhav-kaushal:
Was this tested against PostgreSQL? 

I do not see any format except `text` type being handled as **blob**.

nlomb:
testet this with: mysql  Ver 14.14 Distrib 5.7.13, for Linux.
mysql "text" type does not allow non-printables, so use "blob", so it is also with PostgreSQL

I think. with PostgreSQL type "bytea" should work, see:
http://stackoverflow.com/questions/4386030/how-to-use-blob-datatype-in-postgres

vaibhav-kaushal:
@nlomb Thanks for your reply. I tested that on my end as well and it failed. Going through that link and other resources to find out how to fix this.

nlomb:
@vaibhav-kaushal 
do You use PostgreSQL ?
please post here what works, maybe we need different solutions for other databases as well

spekary:
The error was that was that base64 encoding was failing for some combinations of data, so a completely blank record was being inserted into the database. The PHP manual suggests that base64 doesn't work for everything. This new system is configurable. You can turn off compression and encryption, and as long as you don't have byte data in your session, you can use a TEXT column, but that is a lot of IFs, and it leaves session data, with potentially sensitive information unencrypted in your database. Encryption plus a binary capable database column is really the safest way to go for security.

vaibhav-kaushal:
@nlomb Yes. I use PostgreSQL for a lot of things :) 

@spekary I understand the concern. I am yet to test the `bytea` data type and I hope that one works because if it does not, then we will have to *encode* the data for PostgreSQL. Moreover there is no proper handling of `bytea` type in PostgreSQL adapter in QCubed, which we will have to build.

PR #1078, opened 2016-08-25
bugfix-SaveState2

spekary:
Another attempt at fixing the save state problem. This was is fixing an ajax draw problem. We were exiting prematurely and not calling the SaveControlState or the Form_Exit methods.

### DIFF
--- a/includes/base_controls/QFormBase.class.php
+++ b/includes/base_controls/QFormBase.class.php
@@ -597,22 +597,22 @@
 
 					// Use Standard Rendering
 					$objClass->Render();
-					break;
+
+				// Ensure that RenderEnd() was called during the Render process
+				switch ($objClass->intFormStatus) {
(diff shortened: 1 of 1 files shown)


includes/base_controls/QFormBase.class.php
```diff
@@ -597,22 +597,22 @@ public static function Run($strFormClass, $strAlternateHtmlFile = null, $strForm
 
 					// Use Standard Rendering
 					$objClass->Render();
-					break;
 
-				default:
-					throw new Exception('Unknown Form CallType: ' . $objClass->strCallType);
-			}
+				// Ensure that RenderEnd() was called during the Render process
```
olegabr: needs one more tab to fit in the code around it

spekary: OK. Done.

PR #1079, opened 2016-08-27
Blob Support for PostgreSQL - General and Session Handler

vaibhav-kaushal:
The #1077 PR broke the DbBackedSessionHandler for PostgreSQL. This one re-establishes the support and enables BLOB support in general for PostgreSQL.

### DIFF
--- a/includes/framework/QDbBackedSessionHandler.class.php
+++ b/includes/framework/QDbBackedSessionHandler.class.php
@@ -123,6 +123,15 @@
 			if (!$result_row) // either the data was empty or the row was not found
 				return '';
 			$strData = $result_row['data'];
+
+			if(strstr($objDatabase->Adapter, 'PostgreSql')) {
+				if(function_exists('pg_unescape_bytea')) {
+					$strData = pg_unescape_bytea($strData);
+				} else {
+					throw new QCallerException('pg_unescape_bytea method needed for DbBackedSessionHandler to operate on a PostgreSQL database. Please install the "pgsql" PHP extension.');
+				}
+			}
+
 			if (!$strData)
 				return '';
 
--- a/includes/database/QPostgreSqlDatabase.class.php
+++ b/includes/database/QPostgreSqlDatabase.class.php
@@ -873,6 +873,7 @@
 					$this->strType = QDatabaseFieldType::VarChar;
 					break;
 				case 'text':
+				case 'bytea':
 					$this->strType = QDatabaseFieldType::Blob;
 					break;
 				case 'timestamp':


includes/framework/QDbBackedSessionHandler.class.php
```diff
@@ -123,6 +123,15 @@ public static function SessionRead($id) {
 			if (!$result_row) // either the data was empty or the row was not found
 				return '';
 			$strData = $result_row['data'];
+
+			if(strstr($objDatabase->Adapter, 'PostgreSql')) {
```
olegabr: This should be done in the FetchArray implementation of the Pg adapter

vaibhav-kaushal: IMHO, no. That code is run everywhere! For it to work correctly in FetchArray method in Postgres adapter, the information about the datatypes of each field in the row is to be passed to the method as well - which is too big a task considering the complexity of the rest of the framework.

spekary: Since we don't do runtime analysis of the database, but only do that during codegen, @vaibhav-kaushal is right. Either we need to make this code go through the data model, or we need a special case for postgresql since it does not have good support for BLOB or binary columns.

olegabr: @spekary I understand your point here. Lets look on it from other side:
1. here we obtain data some database
2. we know that the data is a `blob`

It looks like we need some database-agnostic method to obtain typed data. E.g. 

```php
$objDatabase->getValue($strData, QDatabaseFieldType::Blob);
```

It looks like data-conversion layer to be refactored from the DB layer and exposed to the end user maybe.

It is important issue for me, because any `if`-s like that in the codebase are a bad smell of architectural flaws. In this case it is simply a sign that some functionality is over-incapsulated here.

spekary: This is becoming a lot of work for what should be a simple fix. Perhaps there is another way to encode the data so that it will go into a text field. Base64 encoding was failing for some reason.

vaibhav-kaushal: @spekary It would be wrong to say 
> it does not have good support for BLOB or binary columns

It does. It just does not handle them the way MySQL does. 

This is where we need to start working on per-database Data Types - except the basics (INT, FLOAT/REAL/DOUBLE, DATE, VARCHAR), most databases handle the other types differently. Yes, it can't be done *right now*. But we have to move in that direction with time.

olegabr: @vaibhav-kaushal please, create a new issue with a proper problem explanation and assign it to the 3.1 milestone. After that we would be able to merge this PR with a reference to it from the new issue created.

spekary: Re: per database type handling, the problem is that this is hand-written SQL so it will not really help us here. We do not want to read the database schema every time we do a database operation. QQuery is supposed to help us with this, so perhaps we need to turn this into a QQuery operation.

vaibhav-kaushal: @spekary I am saying the same thing. May be you misunderstood me (or I misunderstood you somewhere). For now, we have to let this one go. 

@olegabr I am going to create a new Issue right away.

vaibhav-kaushal: Please see #1083.

spekary:
Is this fixed?

Related to this, I am going to add another request. I don't think we should have Blob types, but rather Binary types. PHP does not care if something is a Blob, VarChar, Char or Text, it treats them the same. However, there are situations where they should be treated differently. Right now, when we see a Blob or Text type, we treat them the same way, both as a Blob. But really, we don't need to know how SQL is internally storing objects, but rather we need to know whether the object is Binary data or string-based data. String data we can sort and create text fields for. Binary data, we should not do any UI creation, but we DO want to make sure we Base64 the data before sending it to json.

PR #1080, opened 2016-08-27
Fixing Installation messages

### LINKED ISSUE #1072: found two minor issues
QInstallationValidator.class.php creates this:
```
Welcome to QCubed!
PHP5 Model-View-Controller framework

This simple wizard will help you configure QCubed for first use. It'll take you just a couple minutes. If you have any questions along the way, feel free to ping us on the support forums, a vibrant community is there to help you all the time. There's also a chat room where you can get help right away.

Here's what you need to do:

    Cache directory (/usr/data/share/www/tegralink/resources/project/tmp/cache/purifier) needs to be writable

Here are commands that can fix several of these issues:

chmod 777 <MY-DOCROOT>/resources/project/tmp/cache/purifier
```
errors: 

1. - should be "PHP5 - PHP7 Model-View-Controller framework"
2. - "ezyang/htmlpurifier" is not installed

vaibhav-kaushal:
Fixes #1072 for now.

Also, I strongly suggest that we put `ezyang/htmlpurifier` in the *require* section in composer.json. 

Not doing so breaks the QCrossScripting with textboxes and is not a pleasant thing, especially for a newbie. Since we are using the library in our code and it is anyways compatible (license-wise), why not install it right away? 

**Additional Note**: We have tested the PHP7 functionality in our organization with the project where we are using the framework. It is not causing any issues right now. Hence, I have added the PHP7 support in Readme and Start pages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Older releases are available from the [downloads archive](https://github.com/qcu
 
 ## What is QCubed?
 
-QCubed (pronounced 'Q' - cubed) is a PHP5 Model-View-Controller framework. The goal of the framework is to save development time around mundane, repetitive tasks - allowing you to concentrate on things that are useful AND fun. QCubed excels in situations where you have a large database structure that you quickly want to make available to users.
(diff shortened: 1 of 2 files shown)


spekary:
htmlpurifier is big and slow. Its also not needed in all situations. If QCrossScripting is broken without it, that should be fixed. Required means absolutely required. We should require as little as possible. Suggested means strongly suggested and you get added benefits from including it. htmlpurifier really fits that category.

PR #1086, opened 2016-08-31
bugfix-EmptySession

spekary:
Fixing a problem caused by saving an empty session while the encryption and compression tags were off. Some other cleanup as well.


spekary:
Don't merge yet. I realized that an empty session means we should delete the record.

spekary:
OK now

PR #1090, opened 2016-09-07
Add script codegen_cli.php

lackovic10:
The goals of this PR is to add a script which enables codegen execution from the cli
- [x] add `/assets/php/_devtools/codegen_cli.php`
- [x] add description to the `_README.txt` file

### DIFF
--- a/assets/php/_devtools/_README.txt
+++ b/assets/php/_devtools/_README.txt
@@ -7,6 +7,8 @@ This directory contains the web-based drivers for Qcubed's development tools:
   activities in a nicely formatted HTML page.  It uses the codegen_settings.xml
   file as the "settings" to use for codegenning.
 
+* codegen_cli.php - the Qcubed CodeGen cli based driver.
+
 * (future tools tba)
 
 Feel free to alter the settings, inputs and/or outputs of any of the drivers
--- a/assets/php/_devtools/codegen_cli.php
+++ b/assets/php/_devtools/codegen_cli.php
@@ -1,0 +1,41 @@
+<?php
+
+set_time_limit(86400);
+
+if (php_sapi_name() !== 'cli') {
+    echo 'Only for command line interface usage';
+    exit();
+}
+
+require __DIR__ .'/../qcubed.inc.php';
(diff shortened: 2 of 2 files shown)


lackovic10:
@spekary can you please add a wip label ?

lackovic10:
thanks @spekary, would it be possible for us to assign labels ?
this PR is ready for review

olegabr:
how we can test it? may be use it in travis test script?

olegabr:
see here: https://github.com/qcubed/framework/blob/beta-3.0/travis/test.php#L34

lackovic10:
are you testing the codegen class somewhere ?

lackovic10:
@olegabr i've checked `codegen.inc.php`
i don't think we should run the codegen again in the travis test, the codegen already seems to be ran there
i suggest merging this, and that codegen itself needs proper tests (if there aren't any already) and those can be added later

vaibhav-kaushal:
I am not sure if Travis will let the tests create files on their system.

vaibhav-kaushal:
@lackovic10 What does this PR actually do? It looks like it runs the codegen via CLI. Is that what it does?

lackovic10:
@vaibhav-kaushal yes, this PR runs codegen from the cli, i've updated the PR description

olegabr:
@lackovic10 please, add a line in the `README` file about motivation to use this tool. After reading this line the user should have an idea in which situation he needs this tool. Thank you!

vaibhav-kaushal:
I think we already have this thing in our _devtools directory somewhere.

lackovic10:
@vaibhav-kaushal looks like there is [codegen.cli](https://github.com/qcubed/framework/blob/beta-3.0/includes/_devtools/codegen.cli)
why does it have the extension `cli` instead of php ?

vaibhav-kaushal:
@lackovic10 I have no idea.

vaibhav-kaushal:
@lackovic10 and @spekary I think this should be in the `project` folder after installation. Which means, it has to be moved to the install directory.

Issue #1088, opened 2016-09-01
Use OpenSSL methods in QCryptography


vaibhav-kaushal:
PHP 7.1 (soon to be released) is going to deprecate (and soon remove) the support for mcrypt. Links: 

  - https://dotdev.co/upcoming-changes-in-php-7-1-76ebea53b820#.i4kcif1lx
  - https://github.com/php/php-src/pull/1996
  - https://wiki.php.net/rfc/mcrypt-viking-funeral

OpenSSL Library is the recommended way to go. 

At present, we are using mcrypt in QCryptography class/file. I believe we should replace that with openssl_* functions.

spekary:
OK. I think this needs to be in an update to 3.0 since that website points out that mcrypt hasn't been updated since 2007 and is basically obsolete.

vaibhav-kaushal:
From what I have been reading in this context, there is this library: https://github.com/jedisct1/libsodium which seems to be gaining momentum slowly, steadily. I have not gone through it much so can't say much but looks promising from what I have read. The pain is - you have to **compile** it. Not all developers are happy with that. Even if you are an experienced developer, to compile it everywhere is an un-necessary pain. 

At the same time it looks like openssl is built into PHP 5.3+ (from what I see on Mac (native+homebrew) and Ubuntu) - which mean no need to compile or install anything! I think we need to confirm if openssl is already built-in and if it is, then we can move ahead with that. 

Also, this time, let's make AES-256 the default encryption, if possible. While Triple-DES is no issue either, AES is stronger and **recommended** (by US Government, and that makes it quite famous; the aim being - the defaults are set to something which is widely accepted (Intel has hardware instructions for AES and no other Cipher) and recommended).

aadityabagga:
Maybe this can be used:
https://paragonie.com/blog/2015/05/if-you-re-typing-word-mcrypt-into-your-code-you-re-doing-it-wrong

vaibhav-kaushal:
From what I have seen so far, openssl extension seems to be enabled by default in PHP with every distribution above 5.3.

Since no one has put in any comments since some time here, I am moving ahead with implementation of QCryptography using openssl functions available in PHP.

spekary:
Yes, please.

vaibhav-kaushal:
Please check the PR. 
#1092 
Closing this in favor of the PR.

PR #1092, opened 2016-09-09
QCryptography Updated

vaibhav-kaushal:
Here is the list of changes done:

  - Remove the mcrypt module functions and the corresponding object being used in the file.
  - Change the default algorithm to AES-265
  - Change the default mode from ECB to CBC
  - Set a default IV for QCryptography to maintain the backward compatibility with simple initializations like `$objCrypto = new QCryptography();`
  - Remove the RandomSource and Mode option and instead allow the user to put in an IV string by his own.
  - Moved the default Cipher name, Key and IV to be used in the configuration file. This is to make sure that subsequent framework upgrades do not override user defaults and the user does not need to override a core class just for the sake of changing defaults of his cipher of choice. 

This PR is for #1088 .

### DIFF
--- a/includes/framework/QCryptography.class.php
+++ b/includes/framework/QCryptography.class.php
@@ -20,8 +20,6 @@
 		 */
 		const IV_RANDOM = 'R';
 
-		/** @var resource Mcrypt algorithm module resource */
-		protected $objMcryptModule;
 		/** @var bool Are we going to use Base 64 encoding? */
 		protected $blnBase64;
 		/** @var string Key to be used for encryption/decryption - used for mycrypt_generic_init */
--- a/includes/framework/QCryptography.class.php
+++ b/includes/framework/QCryptography.class.php
@@ -208,13 +208,4 @@
 		public function __destruct() {
 
 		}
-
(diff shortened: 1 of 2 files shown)


vaibhav-kaushal:
This is working properly in my tests (with my applications, of course). I will merge this PR in case no one else has the time to test. We should include/discuss the changes in this PR when we build the changelog.

olegabr:
The `objMcryptModule` member field and it's usages in `__sleep` and `__wakeup` are useless now.

vaibhav-kaushal:
@olegabr Removed them. Thanks for pointing out. Had not noticed them.

olegabr:
great! just one more fix, please: the field `objMcryptModule` is still there.

vaibhav-kaushal:
Done! 

Sent from a rotary phone.




On Mon, Sep 12, 2016 at 4:05 PM +0530, "Oleg Abrosimov" <notifications@github.com> wrote:












great! just one more fix, please: the field objMcryptModule is still there.

olegabr:
Great job!

Issue #1032, opened 2016-07-21
Whether there is one problem, which is for me with Profiling Information?


kukrik:
I did a little work on the CSS, what you see herein below. It is a style of polite:

<img width="1280" alt="screen shot 2016-01" src="https://cloud.githubusercontent.com/assets/2853034/17041369/61c0bb8c-4fae-11e6-8bb5-dcfd3b52e8fe.png">
<img width="1280" alt="screen shot 2016-03" src="https://cloud.githubusercontent.com/assets/2853034/17041371/61c2dcaa-4fae-11e6-9d76-44c2db7d14ee.png">

I found here the configuration.php inside the line of `'encoding' => 'utf8'` dislike to Profiling Information. It throws the error: MySqli Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SET NAMES utf8' at line 1.

<img width="1280" alt="screen shot 2016-02" src="https://cloud.githubusercontent.com/assets/2853034/17041370/61c1b794-4fae-11e6-952d-0a6002910134.png">

If this line (`//'encoding' => 'utf8'`) to hide it Profiling Information is working properly. But this line is very necessary to save and display the accented characters correctly.

I have to realize that this is a simple mistake.

But I was looking for a solution, have not found the right solution ...

Before you want to ask how you are running Profiling Information. Linux platform on top, or whatever?

I would like to provide improvements in CSS before QCubed 3.0.0.

EDIT: I have a MySQL version 5.7.10.

PR #1098, opened 2016-09-13
Bugfix small adjustments of style

kukrik:
1) Added and modified corepage.css.
2) One mistake removed at styles.css
3) Small tweaks at profile.php and error_dump.php

Reference: https://github.com/qcubed/framework/issues/1032


kukrik:
Too much is placed there, which is not necessary here.

I do a little bit of work around...

PR #1099, opened 2016-09-13
Bugfix small adjustments of style

kukrik:
1) Added and modified corepage.css.
2) One mistake removed at styles.css
3) Small tweaks at profile.php and error_dump.php

Reference: https://github.com/qcubed/framework/issues/1032

### DIFF
--- a/assets/css/corepage.css
+++ b/assets/css/corepage.css
@@ -1,0 +1,144 @@
+/**
+ * Borrowed css from qcodo
+ * Modified according to the need for a qcubed:
+ * profiling
+ * error page
+ */
+
+	body { 
+		font-family: Arial, Helvetica, sans-serif;
+		font-size: 14px;
+		margin: 0;
+	}
+	a { color: #009721;}
+	a:link, a:visited { text-decoration: none; }
+	a:hover { text-decoration: underline; }
+	form { margin: 0; }
+	
+	#container { }
+	#content { padding: 15px 20px 0 20px; margin-bottom: 20px; }
+	
+	#headerContainer {
+		overflow: hidden;
+		padding-bottom: 5px;
+		line-height: 18px;
+		width: 100%;
+	}
+	#headerBorder { }
+	#header {
+		background-color: #333;
+		color: #ffffff;
+		font-size: 16px;
+		overflow: hidden;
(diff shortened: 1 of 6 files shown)


kukrik:
This issue #1032 should not be overlooked. Profiling Information is not working as expected, if the encoding is inside. 

This issue needs to be resolved :(

vaibhav-kaushal:
What changes are these?

kukrik:
We agreed that a picture is better than a thousand words. I show before and after the changes.

**Before:**
before_1

**After:**
after_1

**Before:**
before_2

**After:**
after-2

**Before:**
before_3

**After:**
after_3

My mistake was that I put together two themes. My amendments relate only to the first topic.

Another topic that profiling Information is not working as expected, if the encoding is inside. 

Probably there will be a separate topic for this issue. But it is also an important issue!

Thoughts?

kpirbhai:
@kukrik 
Are you referring to if the profiling output is within the main "Form" block that it doesn't work much of the time? If so, I've been having the problem as well. I've looked at it some but haven't had enough time to really dig in. I put RenderBegin and RenderEnd in the include header and footer respectively right after and before the body tags. I've gotten errors that $j and or jQuery aren't defined when clicking on the profiling output link.

kukrik:
@kpirbhai, maybe I understand your question  correctly. This is another issue that I wanted to talk about. 

Database Profiling Tool does not work as expected when in configuration.inc.php is `'encoding' => 'utf8'` enabled.  It throws the error: MySqli Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SET NAMES utf8' at line 1.

When `'encoding' => 'UTF-8'` is not active, it Database Profiling Tool works as expected.

I think the problem lies elsewhere, not what you thought?!

kukrik:
I'll take back to patch bugfix-encoding ago.  I'm not sure whether this is a bug? My development server shows an error with app, but in production server not the problem.

Do we have to encoding and/or set_charset theme investigate?

The patch bugfix_small_adjustments_of_style can only merge

olegabr:
This PR is pretty clear. Thank you, @kukrik!

PR #1100, opened 2016-09-15
bugfix-QCrypto

spekary:
Fixing to make sure IV's are randomly generated always, and providing a mechanism to allow values to be decrypted without needing to save the IV, but rather providing the IV with the encrypted data. Also added unit tests and documentation.


vaibhav-kaushal:
Looks good. Makes sense. 👍

PR #1109, opened 2016-10-01
add container

lackovic10:
Added a simple container `/includes/container/AbstractContainer`, which needs to be extended and the services need to be defined manually
A usage example can be found in `includes/tests/qcubed-unit/container/Container.php`


lackovic10:
@spekary what do you think about the idea, this simple container works very well for me and it works well with qforms
in my company i've created a main container which extends the `AbstractContainer` and a `TestContainer` which extends the main container and defines services used only in tests

spekary:
I like the direction of this. It supports a few design patterns I think we should move to: microservices (ish), middleware possibilies too. However, there are some things still to flesh out, like multiple containers, and whether to use the PHP SPL to do the container management. @olegabr I would like your thoughts on this, but I am thinking we should push this to a 3.1 branch and move 3.0 to rc status now.

lackovic10:
would it be possible to make it as a separate component and put it into the folder
`src/QCubed/Component/Container` ? i'd like to setup namespaces and make it compatible with the currently pending [psr container standard](https://github.com/php-fig/fig-standards/blob/master/proposed/container.md)
what do you mean by multiple containers and using PHP SPL ? i assume by spl you mean standard php library ?

spekary:
Can you make this pull request against v3.1? Then we can move forward with it as is I think.

lackovic10:
yes, i will make the PR against 3.1
from what i understood, 3.1 is going to use namespaces so i'd like to update that and maybe some other updates before merging
also feel communication on these PR's is quite slow, can i send some general questions on email?
my email is lackovic10@gmail.com

spekary:
4.0 is the version with namespaces and decoupling of the framework. 3.1 will contain some intermediate efforts that will help us get there.

spekary:
For 4.0, these are going to be broken out into separate repos corresponding to the namespaces. What do you want to work on first, and I will create a repo for it here?

lackovic10:
ok i will make this on 3.1. i will also add one update, the container should not be serialized, i've had problems with unserialize, some might have been caused by some services, the problems with unserialize are so hard to debug, we have only errors on production can't really reproduce locally

lackovic10:
as for the version 4.0, i'd like to attempt to decouple the code generator and the query builder and the database connection
i think that should be all in one library but i will think about this, and will do some reading
if you make  a separate repository, can you give me access to merge PR's for that repository ?

spekary:
If the container has any state information in it at all, it will have to be serialized. Alternatively, it could save its state in the session.

lackovic10:
no, i don't want to save the state, on `QAjaxAction` a new container should be created
i will just add a sleep method
```
    public function __sleep()
    {
        // serialize all variables but the container
        return array_diff(array_keys(get_object_vars($this)), array('container'));
    }
```

vaibhav-kaushal:
I had not paid attention to it yet. But after looking at it, though your ( @lackovic10 ) long term plan is not really clear, it does seem to be a nice addition. Middlewares are necessary and there had to be an order to that chaos. This can be one order.

lackovic10:
i will add this on 3.1 as soon as i find a bit of time
i would also like to generate a container class in the project directory so that the user could have  a container already
can someone point me where to add that ?

lackovic10:
closing this one will open a new PR

PR #1116, opened 2016-10-10
code coverage for qcubed

olegabr:
first step to enable code coverage for qcubed


olegabr:
After merge of this PR we should have a picture of how each new PR influences the code coverage and other code metrics.

PR #1121, opened 2016-10-26
feature-DbCodegenEnhancements

spekary:
- Adds getters and setters.
- Routes __set and __get request to getters and setters if they exist.
- Update command only updates the fields that were changed rather than all the fields in the table.
- Getters will make sure that the data was retrived by the last query. In other words, if the data was not retrieved because of a QQ::Select clause, and you then try to read the data, it will throw an exception.
- Update provides more granular ability to watch for updates. You can now detect directly if an item was added, or if a particular field was updated.
- copy method will make sure id and flags are reset
- Will detect default boolean values and set those if appropriate
- New model objects will now be initialized using database defaults by default (this is configurable in codegen_settings.xml
- Objects are now json serializable, and QQ::Select will determine what actually gets serialized.


olegabr:
@spekary, great addition!

PR #1132, opened 2016-11-15
fixes #1131

### LINKED ISSUE #1131: Installation process does not work
I've tried both the composer installation and just downloading the qcubed and running the included installation procedure and neither one of the works at all.  I give up!

olegabr:
`qcubed/framework` replaced with `qcubed/qcubed`


vaibhav-kaushal:
We will have to do this in plugin repos as well.

PR #1137, opened 2016-11-20
bugfix-php72

spekary:
PHP 7.2 is now in the nightly build and is causing Travis to fail. The problems are two things:

count() does not handle a null value as an array. This uncovered a problem in code generation in the column analyzer.

Also, DateTime::setTime added a microsecond parameter.

### DIFF
--- a/includes/base_controls/QListItemManager.trait.php
+++ b/includes/base_controls/QListItemManager.trait.php
@@ -28,7 +28,11 @@
 		 */
 		public function AddListItem(QListItemBase $objListItem) {
 			if ($strControlId = $this->GetId()) {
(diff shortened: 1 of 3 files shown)


vaibhav-kaushal:
I saw the PR and I think it's good. But I would suggest that we wait until the version is released as stable. That way we can be sure that these changes need to stay!

spekary:
Travis build will not complete without them. Besides, they are truly bugfixes. The codegen stuff was not correctly determining the Unique value because it was trying to see if there were other columns that were part of the primary key, before the other columns were defined. That is why the array was null when we were trying to count it. And the setDate function has changed in 7.2, so that had to change anyways.

PR #1138, opened 2016-11-21
Add container

lackovic10:
[PR 1109](https://github.com/qcubed/qcubed/pull/1109)

Added a simple container `/includes/container/AbstractContainer`, which needs to be extended and the services need to be defined manually
A usage example can be found in `includes/tests/qcubed-unit/container/Container.php`

### DIFF
--- a/assets/php/examples/basic_qform/dependency_injection_container.php
+++ b/assets/php/examples/basic_qform/dependency_injection_container.php
@@ -1,0 +1,114 @@
+<?php
+
+require_once('../qcubed.inc.php');
+require('../includes/header.inc.php');
+?>
+<div id="instructions" class="full">
+    <h1>Dependency Injection Container</h1>
+
+    <p>
+        According to Fabien Potencier, the author of Symfony, a Dependency Injection Container is an object that knows how to instantiate and configure objects. And to be able to do its job, it needs to knows about the constructor arguments and the relationships between the objects.
+    </p>
+    <p>
+        <h3>Related Links</h3>
+        <a href="http://fabien.potencier.org/do-you-need-a-dependency-injection-container.html">
+            Fabien Potencier - series of articles on Dependency Injection
+        </a><br />
+        <a href="https://www.martinfowler.com/articles/injection.html">
+            Martin Fowler - Inversion of Control Containers and the Dependency Injection pattern
+        </a>
+    </p>
+    <p>
+        QCubed adds a simple <i>includes/container/AbstractContainer.php<i>, which needs to be extended and the services need to be added manually to the <b>AbstractContainer</b> implementation.<br />
+        An example implementation:
+    </p>
+    <pre>
+        <code class="php">
+        class Container extends AbstractContainer
+        {
+            /** @var array $services */
+            protected $services = array(
+                'logger' => null,
+                'repository.user' => null,
+                'manager.registration' => null,
+            );
+
+            /**
+             * @param string  $id
+             * @param boolean $newInstance
+             *
+             * @return Logger
+             */
+            protected function getLoggerService($id, $newInstance)
+            {
+                if ($newInstance || !$this->services[$id]) {
+                    $this->services[$id] = new Logger();
+                }
+
+                return $this->services[$id];
+            }
+
+            /**
+             * @param string  $id
+             * @param boolean $newInstance
+             *
+             * @return UserRepository
+             */
+            protected function getRepository_UserService($id, $newInstance)
+            {
+                if ($newInstance || !$this->services[$id]) {
+                    $this->services[$id] = new UserRepository();
+                }
+
+                return $this->services[$id];
+            }
+
+            /**
+             * @param string  $id
+             * @param boolean $newInstance
+             *
+             * @return RegistrationManager
+             */
+            protected function getLoggerService($id, $newInstance)
+            {
+                if ($newInstance || !$this->services[$id]) {
+                    $this->services[$id] = new RegistrationManager(
+                        $this->getService('repository.user', array(), $newInstance),
+                        $this->getService('logger', array(), $newInstance)
+                    );
+                }
+
+                return $this->services[$id];
+            }
+        }
+        </code>
+    </pre>
+    <p>Usage Example:</p>
+    <pre>
+        <code class="php">
+        class RegisterForm extends QForm
+        {
+            protected $btnRegister;
+
+            protected function Form_Create()
+            {
+                $this->btnRegister = new QButton($this);
+                $this->btnRegister->AddAction(new QClickEvent(), new QServerAction('btnRegister_Click'));
+            }
+
+            protected function btnRegister_Click()
+            {
+                $this->container->getService('manager.registration')->register();
+            }
+        }
+
+        $container = new Container();
+        RegisterForm::Run('RegisterForm', 'templates/register.tpl', null, $container);
+        </code>
+    </pre>
+    <p>
+    If a new dependency is added to the <i>RegisterManager</i> class, it needs to be added only in one place, in the Container.
+    </p>
+</div>
+
+<?php $mainPage = true; require('../includes/footer.inc.php'); ?>
--- a/assets/php/examples/includes/examples.inc.php
+++ b/assets/php/examples/includes/examples.inc.php
@@ -65,13 +65,14 @@
 			self::AddCoreExampleFile($intIndex, '/basic_qform/intro.php Hello World Example');
 			self::AddCoreExampleFile($intIndex, '/basic_qform/state.php QForms: Stateful, Event-Driven Objects');
 			self::AddCoreExampleFile($intIndex, '/basic_qform/process_flow.php Understanding Process Flow');
+			self::AddCoreExampleFile($intIndex, '/basic_qform/dependency_injection_container.php Dependency Injection Container');
 			self::AddCoreExampleFile($intIndex, '/basic_qform/calculator.php Calculator Example');
 			self::AddCoreExampleFile($intIndex, '/basic_qform/calculator_2.php Calculator Example with Validation');
 			self::AddCoreExampleFile($intIndex, '/basic_qform/calculator_3.php Calculator Example with &quot;Design&quot;');
 			self::AddCoreExampleFile($intIndex, '/basic_qform/listbox.php * Introduction to QListControl');
 			self::AddCoreExampleFile($intIndex, '/basic_qform/hlist.php * Generating Html Lists');
 			self::AddCoreExampleFile($intIndex, '/basic_qform/textbox.php * Introduction to QTextBoxControls');
-			
+
 			$intIndex++;
 			Examples::$Categories[$intIndex] = array();
 			Examples::$Categories[$intIndex]['name'] = 'Basic AJAX in QForms';
@@ -91,8 +92,7 @@
 			self::AddCoreExampleFile($intIndex, '/events_actions/other_actions.php Other Client-Side QActions');
 			self::AddCoreExampleFile($intIndex, '/events_actions/event_propagation.php Controlling Event Bubbling');
 			self::AddCoreExampleFile($intIndex, '/events_actions/event_delegation.php Event Delegation');
-			self::AddCoreExampleFile($intIndex, '/events_actions/synchronizing.php Blocking Unwanted Events');
-			self::AddCoreExampleFile($intIndex, '/events_actions/low_priority_js.php JavaScript Priorities');
+			self::AddCoreExampleFile($intIndex, '/events_actions/low_priority_js.php JavaScript priorities');
 			self::AddCoreExampleFile($intIndex, '/events_actions/watcher.php Automatic Refreshing of Controls');
 			self::AddCoreExampleFile($intIndex, '/events_actions/control_proxy.php Using QControlProxies to have Non-QControls Trigger Events');
 
@@ -154,7 +154,7 @@
 			self::AddCoreExampleFile($intIndex, '/other_controls/jq_example.php * Server-side wrappers for all jQuery UI Controls');
 			self::AddCoreExampleFile($intIndex, '/other_controls/js_return_param_example.php Post data back to the server from jQuery UI controls');
 			self::AddCoreExampleFile($intIndex, '/advanced_ajax/jquery_effects.php JQuery Effects');
-			
+
 			Examples::$AdditionalCode[__VIRTUAL_DIRECTORY__ . __EXAMPLES__ . '/advanced_ajax/dialog_box.php'] = array('CalculatorWidget.class.php','CalculatorWidget.tpl.php');
 
 			$intIndex++;
@@ -198,7 +198,7 @@
 			self::AddCoreReferencedFile('/communication/soap.php', 'example_service.php');
 			self::AddCoreReferencedFile('/other/parsing.php', 'bb_parser.php');
 			self::AddCoreReferencedFile('/communication/advanced_intl.php', 'sample_translator.class.php');
-			
+
 
 			$intIndex++;
 			Examples::$Categories[$intIndex] = array();
@@ -267,7 +267,7 @@
 		private static function AddCoreExampleFile($intIndex, $strExampleFileName) {
 			array_push(Examples::$Categories[$intIndex], __VIRTUAL_DIRECTORY__ . __EXAMPLES__ . $strExampleFileName);
 		}
-		
+
 		private static function AddCoreReferencedFile($strExampleFileName, $mixReferencedFile) {
 			if (!is_array($mixReferencedFile)) {
 				$mixReferencedFile = array($mixReferencedFile);
@@ -280,7 +280,7 @@
 			$categories = Examples::$Categories;
 			for ($intCategoryIndex = 0; $intCategoryIndex < count(Examples::$Categories); $intCategoryIndex++) {
 				$objExampleCategory = Examples::$Categories[$intCategoryIndex];
-				
+
 				for ($intExampleIndex = 0; $intExampleIndex < count($objExampleCategory); $intExampleIndex++) {
 					if (array_key_exists($intExampleIndex, $objExampleCategory)) {
 						$strExample = $objExampleCategory[$intExampleIndex];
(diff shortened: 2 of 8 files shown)


lackovic10:
@spekary can this be merged? i wanted to add examples but go no feedback. examples can be added later in a new PR

kukrik:
@lackovic10, be sure to add examples! Be sure to get others to understand, I remain a little unclear...

lackovic10:
@kukrik can you please point me to how can i add them? i would be happy to chat on slack next week about it

kukrik:
Now I understand, that I don't have the rights to do anything.
For example, if that's status will to be mine and i must tell, now i don't want, because i am not yet special for it. 

Example to add /vendor/qcubed/qcubed/assets/php/examples/... And to add one line in .../includes/examples.inc.php.

@spekary, @olegabr, @vaibhav-kaushal is owners, if they accept, then merged it.

lackovic10:
thanks @kukrik i will add an example

vaibhav-kaushal:
@lackovic10 I really need to understand what this PR does. From what I can understand, the functionality is as follows:

- All Forms (QForms) will have something called a container. 
- These containers are arrays which contain instances of classes. 
- One can ask for a new class from a method which acts like a service. 
- The service can either search for an existing instance in the container or can create a new one and return it (while also saving it in the array), depending on what was requested. 

So the container serves as a bag of things in it of which, one can be returned. Please let me know if I am wrong.

lackovic10:
@vaibhav-kaushal a container is an object not an array, and yes, it's like a bag which manages creation and retrieval of services. you can ask for a new instance with this implementation, i don't use that feature often, mostly just returning the same instance of a service.
the job of the container is to manage dependencies of objects
here are some articles on dependency injection (if i remember you said you are familiar with it)
http://fabien.potencier.org/what-is-dependency-injection.html
https://www.martinfowler.com/articles/injection.html
also feel free to message me on slack if you want to discuss

lackovic10:
@spekary @vaibhav-kaushal i've added an example, please double check

vaibhav-kaushal:
If this is working and does not break old installations, then I do not think this should be a problem.

vaibhav-kaushal:
@spekary Please have a look at this and tell us if it should be merged into the master or not.

spekary:
I do not like the idea of creating a new dependency for the core when we are adding optional functionality that is not required. We are trying to move to a more modular system, not less modular.

It seems to me like the entire thing could be done as an add-on, which for now happens through our plugin system. The plugin system works well for this right now since it also lets you namespace the add-on, and you can add your own examples. This would let you develop it at your own pace without waiting for us to review and approve changes. 

Also, I think a better place for this would be the QApplication class rather than QForm. The QApplication is the global object, encapsulating things that are consistent across the entire system.

My suggestion here is to use the QBootstrap plugin as a model. It requires a little installation work to override the QControl class, but that process means that QControlBase can stay the same and all functionality is overlayed on top of the system.

I suggest you document how to add a line to the QApplication class (which is totally under the power of the user to modify), and instantiate a custom Container class. They can then install the rest of it using the plugin system, you get the benefit of the composer autloader, and again you can essentially be the owner of the plugin.

spekary:
To be more specific about the way this could be done:

1) Add a line to the QApplicaiton class in the install/project/configuration directory, and comment the line:
```
// Uncomment the following line to add a Container class. See the QContainer plugin for more info.
// \QCubed\Plugin\QContainer::Initialize();
```
2) Your plugin should have a QContainer.class.php file in the "install" directory that will be a stub or example of what you envision your container registry file should look like. This file will get installed into the project/includes/plugins directory, and will be the place for user definable and configurable functionality. It should have in it:
```
public static function Initialize() {
}
```
Which will be the place your services get registered and instantiated. Note that this will get run every time to program enters, whether an Ajax or Server call.
3) Your plugin should have an includes directory that will contain all of your base and abstract classes, things that are not user configurable.
4) Put examples in your examples directory, and reference them in your composer.json file (see the QBootstrap composer.json file)
5) Create a readme to document how to install and use it.
6) We will add it as a repository to the project and make you an owner.
7) We should then add it to the Packagist site so anyone can install it.

lackovic10:
i like the idea of a modular approach. i don't think that you need a new repository for a new component. check how symfony does it https://github.com/symfony/symfony/tree/master/src/Symfony/Component it does not seem to me that they have a separate repository for each component, i'll look into it.
i don't like the idea of adding anything to `QApplication` and i don't like the idea of `QApplication` itself. the point of the container is to enable managing dependencies and get rid of `static` methods which are the main cause of spaghetti code in my opinion and qcubed is full of them
i will look at the plugin system when i find a bit of time

spekary:
> separate repository for each component
Symfony is a bit of a monster at this point, but the idea there is that those are all core components, things that the Symfony community is supporting. We will move to this model, and we have a discussion going on what it should look like, but I don't see us creating something that big.

Regarding static code, I don't think there is anything inherently spaghetti'ish about static code. Spaghetti is a lack of organization, and either starts with poor design, or creeps in over time, which all projects suffer from eventually and need to be refactored from time to time. Dependency injected code can certainly become disorganized too. QCubed has some, but I wouldn't say its everywhere. QApplication has it because its part of the bootstrap process, and the Models have it for object access and instantiation, to separate out code that creates objects from code that operates on objects.

All code needs a bootstrap process, and will inherently then have some kind of static and global space. You can choose to minimize that, but you still need some. QApplication happens to be a convenient container for globally accessable code, objects and settings. Looking at how you are using dependency injection (for logging, user wide settings), and what is inserted there, it seems this is more an application wide thing, rather than form specific. 

The main differentiating thing is whether these objects have internal state that needs to persist alongside the form data. If so, they should be in the form object and will then automatically get serialized. If not, they should be in the QApplication object, since that is recreated on every application entry. One of your modules appears to be a user settings module (which is great!) I assume that has its own persistance mechanism (_SESSION data, or a user data table?). Since the persistance mechanism is not the formstate mechanism, that also indicates it should be part of QApplication.

QApplication is just the entry point. Your static QContainer::Initialize function can create a global object if you think that is best. You might create one more static function to access the global object (QContainer::getService($strServiceId) ?), and then from there you are in to objects and no more static  functions.

lackovic10:
there are lots of articles why static methods are bad, here is a list i quickly put together i didn't read them at this point:
https://r.je/static-methods-bad-practice.html
http://blog.schauderhaft.de/2013/07/07/why-static-is-bad-and-how-to-avoid-it/
http://verraes.net/2014/06/when-to-use-static-methods-in-php/
http://stackoverflow.com/a/752805/1924250
https://blog.goyello.com/2009/12/17/why-static-classes-are-evil/

if you have `QContainer::getService($strServiceId)` you loose the whole point of the container. The container should be accessible only in Controllers (which don't exist in qcubed so far) and in QForms (which are like controllers to me).
the point is NOT to be able to do
```
class User
{
    public static function Register()
    {
        QContainer::getService('logger')->log();
        QContainer::getService('password_manager')->generate();
        QContainer::getService('email')->send();
        ...
    }
}
```
it needs to be
```
class RegisterManager
{
    private $passwordManager;
    private $logger;
    private $emailManager;

    public function __construct(PasswordManager $passwordManager, LoggerInterface $logger, EmailManager $emailManager)
    {
        $this->passwordManager = $passwordManager;
        $this->logger = $logger;
        $this->emailManager = $emailManager;
    }

    public function register(User $user)
    {
        $this->logger->log();
        $this->passwordManager->generate();
        $this->emailManager->send();
        ....
    }
}
```
```
protected function btnRegister_Click()
{
    $this->container->getService('register_manager')->register(User $user);
}
```
i hope you get the point from my code it's just something i threw together quickly it's not production code.
try and go through your `User.class.php` and count how many different classes (dependencies) are you using.
i've worked on three relatively big projects with qcodo/qcubed and all had exactly the same problem.
one was rewritten completely to symfony, one is not really being developed a lot, the code is quite a mess, and the project i'm currently working on uses a `Container`.
we are currently removing qforms from the frontend and the idea is to leave the admin on qforms.
the container is also helping the code with qforms and for the code without qforms i don't even have to mention, i can't imagine how it would work without a class to manage the dependencies.
the project of rewriting is not yet close to finished but we are a decent way into it and so far it has worked out great for us. i can say more once we finish.
The solid principles (https://en.wikipedia.org/wiki/SOLID_(object-oriented_design)) are something that i'm aiming for, i think all of the modern frameworks are based on them.

spekary:
I understand your points. There are a lot of issues at play to unwind.

QForms are not controllers from the application standpoint. From the "Single responsibility principle", trying to make QForms into application wide controllers would be bad. 

A QForm is an abstratction of a single html page with a single html form tag. QForms make a ton of assumptions and put limitations on what you can do with a page, but they work well for many kinds of applications. They specifically are designed to allow OOP style programming within the HTML POST lifecycle by remembering and restoring the state of the controls within the form, and mirroring that state to PHP objects. With the emergence of javascript frontend frameworks, this style of programming is passing away, but for now, its still useful for certain kinds of projects. Think of a QForm as an encapsulation of a representation of the QControl objects on the screen. 

QCubed was never designed to be a pure OOP implementation. It was designed to allow OOP within the confines of the Web at a time when PHP and Web technologies were very immature and limiting. It does some things poorly, but still manages to work for certain kinds of projects.

The main problem with QForms I think is connecting user actions with business logic. QCubed's design paradigm is that the database is the source of truth, which makes putting business logic in the Models and the ModelControllers a good way to handle the problem. I realize not all applications work that way, its just the way QCubed appears to be designed and responds well to applications that want to work that way.

Some have proposed a pub/sub messaging mechanism to make it easier to decouple user actions from business logic,  which I think can work well in some situations. 

You can still accomplish what you want through a plugin by giving your users the following class:
```
namespace \QCubed\Plugin\QContainer;

class Form extends \QForm {
   private $container;

   public function __construct (Container $container) {
       parent::__construct();
       $this->container = $container;
  }
}
```

And then just tell users to use QContainer forms instead of QForms. Or, they could always just modify their own copy of QForm to do that. At least one thing QCubed has done right is to provide enough scaffolding that customizing the framework is fairly easy to do.

spekary:
I have been thinking more about this issue. First of all, I want to be clear that I like your idea. I am just wrestling with the current implementation. 

This issue actually came up for me as I have been doing some Android programming, and I became increasingly unhappy with the singleton design pattern that seems to proliferate in the Java community. I opted for something sort of similar to a container -- a custom application object with all action dependencies as member objects of the application object, and the application object a member of the action object. Its not really a generic services container, but it serves some of the same purposes for code encapsulation. There are some parallels with QCubed, as andoird Actions are very similar to our QForms, and they both have these sort of pseudo application objects.

The problem as I see it is that QCubed really is not designed currently to be dependent on a dependency injection methodology. Perhaps in the future, but not today. As you correctly noted, there are way too many static functions and objects. Part of that was simple history, and part of that was adhering to the philosophy of making static functions whenever a function did not impact an object's member variables, and using an object as a namespace before PHP even had namespaces. So, making this API move that requires all developers to now include the PSR-7 file can't work. We need an alternative.

From a dependency perspective, the QForm is actually very dependent on the QApplication object. One huge problem that you pointed out is that its kind of a pseudo object -- a static abstract class with a bunch of public static member variables. To REALLY do what you want to do, I think the following kind of code could work:
```
$app = new QApplication();
MyForm::Run('MyForm', 'my-template', $app);
```
Then, in your own application object you can create a container object and make that available to your forms. The QApplicationBase should be rewritten so that all those static variables are pulled in to it as member variables, and then there should be public accessors. The application object would then also serve another purpose -- that of a Context object, very similar to the Context object that Google is now using in Go code, which encapsulates all the http router variables and parameters that for us get sucked out of the $_SERVER PHP global.

This application object could then easily be modified for debug and release versions, and to mock various parts of it for testing. 

One tricky part of this is remembering that the application object is designed to be instantiated in both web server mode and command line mode. _prepend.inc.php_ will need to be rewritten a bit, but some things there can be folded in to this new application object. There currently seems to be little logic in what gets put in prepend.inc.php, and what gets put in QApplication. This would all be a major breaking change, so definitely a 4.0 level thing. I will add it as an issue.

If you are not yet willing to commit to rewriting QApplicationBase, then I think it might be best to make your container a static global for now. The static global issue, and having a dependency injection container, are really two separate issues anyway, and one we don't have to solve all at once. 

Also, you don't have to make it a QCubed plugin. Instead, just make your own repository in your own github space, add it to Packagist, and post it in the Wiki so others can use it. I am reworking the docs and Wiki to make easier for user contributions to be added, but feel free to post a description there. Then you have complete control of feature.

spekary:
I am going to close this issue, as I think its best solved by you creating your own Packagist repository and just telling everyone how to use it.

lackovic10:
sorry i was on vacation. i'll check your comments and add my input to this and other issues as well!

lackovic10:
I appologize for the long post

QApplication::$objCacheProvider, i think this does not exist in qcodo, looks to me like it should definitely part of the ORM library.
I see a lot of things that belong to the request and/or to the response, like: $EncodingType, $ServerAddress, $ContentType, $PathInfo, $ScriptName and so on. So why don't you just add a PSR-7 implementation and inject the request object:
```
MyForm::Run('MyForm', 'my-template', \GuzzleHttp\Psr7\ServerRequest::fromGlobals());
```
I know you are rebuilding the QForm object along with the state on post. At my first thought i think it's logical to update the request somehow for the QForm which has been rebuilt from the serialized state. Then you could mark all the stuff which you have in the request as deprecated, and people can start replacing them with data from the request object. That would mark part of the `QApplicationBase` class as deprecated, and could be removed in the next versions. I saw you've extracted the orm stuff, i didn't check the repository yet, would like to try it. I recommend running a cs fixer on it. We are using `friendsofphp/php-cs-fixer`. It's not working good with comments, but maybe we have something misconfigured, didn't look thoroughly.
As i said i did not check the orm repository much so far, but i think that the database connections, the query builder, the orm objects instantiation (`InstantiateDbRow`) need to be decoupled from the orm objects which would basically mean getting away from the active record pattern i think. I have spent some thinking about this recently and am planning to put in more effort in the following months if possible. Reading about the active record pattern [here](https://en.wikipedia.org/wiki/Active_record_pattern), i see that the orm objects (entities) have `save()`, `update()` and `delete()`, so i'm not sure why qcodo generates methods to fetch data which are all static in these classes? I think that lead to putting all the methods for interacting with the database (which are static), as well as all the business logic into these entity classes on all three qcodo/qcubed projects i've worked on. Haven't worked with other active record implementations but based on my experience, this is very bad practice. If you don't want to move away from the active record pattern a possible approach could be passing the connection to the `save()`, `update()` and `delete()` methods. And the load methods and the `InstantiateDbResult()` should go out of entity classes. All this breaks backwards compatibility, actually all my ideas about decoupling qcubed code and getting rid of the static methods are not backwards compatible. Not sure if it's even possible, since the idea is to replace static calls like `QApplication::Redirect()` with something not static. The method `QApplication::Redirect()` could be moved out of the QApplicationBase class easily. `CloseWindow()` as well...
The direction where i'm thinking at the moment about the orm is to completely get away from the active record pattern, to use repositories for database access and update functions, so that's my current recommendation, but i haven't thought this over thoroughly enough. I'm planning to put more effort into it and will be happy to share my ideas.
To get back to the Container, you said `The problem as I see it is that QCubed really is not designed currently to be dependent on a dependency injection methodology`. I agree that QCubed is not designed to depend on a container, but i didn't add the container to make QCubed dependent on it, i've added the container to help extract business logic from data classes and from QForms. A simple example i think i've already pointed out: registration. A website can have multiple ways of registration: a registration form on the website, a form for admins to easily add new users, a tool for batch registration which reads data from a csv file and so on... From my experience with qcodo, this was usually handled with
```
class User extends UserGen
{
    public static function Register(array $data)
    {
        $user = self::LoadByUsername($data['username']);
        if (!$user) {
            throw new Exception('Existing username');
        }

        $user = new User();
        ....

        UserFriend::AddFriends($user, QApplication::GetLoggedInUser());
        Logger::LogRegistration($user)
        ....
    }
}
```
or, even worse, almost the same code copied in three places.
Using services, which are defined in the container helps to at least keep the business logic or part of it decoupled.
```
class UserRepository
{
    public function loadByUsername($username)
    {
        return User::LoadByUsername($username);
    }
}
class FriendManager
{
    public function addFriends(User $user1, User $user2)
    {
        .........
    }
}
class RegistrationManager
{
    private $logger;
    private $userRepository;
    private $friendManager;

    public function __contruct(Logger $logger, UserRepository $userRepository, FriendManager $friendManager)
    {
        $this->logger = $logger;
        $this->userRepository = $userRepository;
        $this->friendManager = $friendManager;
    }

    public function addNewUser(User $userRegistrator, array $data)
    {
        $user = $this->userRepository->loadByUsername($data['username']);
        if (!$user) {
            throw new Exception('Existing username');
        }

        $user = new User();
        ....

        $this->friendManager->addFriends($user, $userRegistrator);
        $this->logger->logRegistration($user)
        ....
    }
}
```
In no way does this harm QForms or QControls, just makes the business logic code clean and much easier testable.
Also, you've mentioned that QForms are not Controllers. I agree, but i don't see a reason why QCubed should not support controllers as well?
For example, in my current project, i've added Controllers and am using them on frontend, while the admin is on QForms.
And the container is injected into the constructor of controllers.
```
protected $container;

public function __construct(ContainerInterface $container)
{
    $this->container = $container;
}
```
Minor thing i noticed, some code style updates could help readability. Many places like:
```
if ($condition1) {
    // do something
} else {
    // do something else
}
```
could be
```
if ($condition1) {
    // do something

    return;
}

// do something else
```
Getting rid of nested conditions helps readability a lot i think.

spekary:
I am running php-cs-fixer on the new code. Also, PHPStorm has a built in fixer too, which I am also running, and I have also created an additional fixer in the tools directory to change cases to camelCase for functions and UPPER_CASE for constants. There is also an on-going tool I am working on that will gather all API changes, and will process old code and change it to the new format. It can't do everything (in particular, fixing constants that are referred to from parent contexts), but its catching about 99% of things.

Regarding ORM changes, I am moving slowly on any dependency injection at this point for this time around. The goal is to get current code running with the new namespaces, and in the process break out as much code as possible into self-contained units.

When possible, I am creating singletons to manage services. This seems to be working, and isn't entirely incompatible with a dependency container. But it is making it easier to see the different parts and how they work together.

There are opportunities to create entirely standalone libraries here. The Cache library is going to become a separate library, and simply a PSR-16 implementation of various caching backends. The i18n has already been done as standalone. The ORM is standalone at this point, and now relies on a singleton Database Service to provide database connections. We are working on a standalone configuration library too.

The QApplication object is going to be broken out into separate logical objects. An HtmlContext object for one is coming. Translations is already broken out. Whatever else I see in there that can be modularized I will. We do need a place for various utility functions, like Redirect though.

Regarding ActiveRecord pattern, it all depends on how you use it. I think as much business logic as possible should be pushed down to the ORM level. If you want your database to be consistent in a large team, as much as possible you should not have the team accessing the database directly, but rather create a functional API layer for the majority of the team to work with, and that layer manages the interaction between data objects and the database itself. Right now, that has to happen by convention. If you want something really enforceable in a team, I suppose only a kind of microservice layer that completely removes direct access to the database would be needed, but that is currently beyond the scope of the ORM. 

The primary reason for the static methods are to serve as various constructors of database objects. You need a constructor somewhere, and constructors are always static. Teams do have to be careful about how they build out access to the ORM though. You are right that the pattern we are using lends itself to bad design. The generated forms code has too much direct knowledge of the database internals (in particular, the limit and order by clauses required to make tables sort). That is a difficult problem to solve though. 

Any project needs a certain amount of architectural self control, and the model objects, if used correctly, can provide a place to implement an API between the database and the application that hides database internals and improves database consistency. But it takes discipline to do that. As they are now, they do not enforce anything.

However, if you have thoughts on a different pattern, lets hear it. I don't know of many other general purpose ways to access a database. Developers are always free to build custom API's with direct calls to SQL, but most don't want to do that.

> but i think that the database connections, the query builder, the orm objects instantiation (InstantiateDbRow) need to be decoupled from the orm objects 

Yes I agree, but it doesn't necessarily mean getting away from the active record pattern. I have discussed this in other posts, but my thinking is to improve the QQuery objects so that they serve as a kind of query language. The QQuery object is passed through an interface to the database adapters, and the adapters will then do the query, build the requested objects, and pass them back. This allows for a great variety of swappable database back ends, middleware (caching, authentication, mocking, etc.), and even microservice implementations.

PR #1139, opened 2016-11-21
feature-MailToUrlWithName

spekary:
Added a few features to the mailto url, including specifying an alternate name.


vaibhav-kaushal:
The change looks cool. Should merge.

PR #1149, opened 2017-01-12
Basic Redis Cache provider built

### LINKED ISSUE #1144: Redis as Caching Backend
I have been using Memcached backend built into QCubed. However, I believe that there are going to users who will need more than that. I, myself have run into a situation where I need to use Redis as the caching backend. However, Redis support is not available in QCubed yet. What do you guys say about using Redis as a caching backend?

Also, we are not using the 'expire-time' parameter for our caches. Both Redis and Memcached allow you to set a timespan after which the cached object will expire. This is pretty useful but unavailable in QCubed. We need to modify the signatures of the `Set` methods in there.

vaibhav-kaushal:
Basic Redis Cache provider built. I have used this in my project and is working fine. 

This Cache Provider has an additional setting for "TTL". Unlike other Caching systems which we have used till now within QCubed do not *persist* data. They use LRU mechanism and are definitely cleared after server restarts. While there is a way to use Redis as a LRU cache (like memcached), the default behavior of redis is to act like a persistent KV store. Also, not everyone is going to run redis as a LRU cache. Their redis installation may not use LRU settings, and in such cases depending on the server's configuration can be troublesome (can fill disk and keep old objects in cache forever). For such cases, the server should clear the contents automatically after some time. I assume caching things for an hour is acceptable. If the user wants to disable/tune this behavior, he can, by editing the configuration file. This behavior is built into Redis and is reliable as well as efficient.

@spekary and @olegabr Please install a Redis server and test it. Also suggest the tests that we need to write for enabling this.

Additionally, I am introducing the 'docs' folder in our top-level directory to make sure that the developer always has enough documentation with him and collaboration around documentation is easy and anyone can contribute to documentation. 

Fixes #1144

### DIFF
--- a/docs/Caching/Redis.md
+++ b/docs/Caching/Redis.md
@@ -5,7 +5,7 @@ Redis is a key-value style data-structures store. The key length as well as the 
  Performance of Redis depends on whether it is running in persistent mode (saving data to disk) or in-memory mode (all contents are within the specified memory limits). Default configurations run in persistent mode. However, depending on use-case (data structures stored, amount of data stored etc.), performance of Redis can be between "a little slower" to "just a little faster" than memcached. However, implementation of Redis is way more powerful because of the data-structures implementation.
  
  ## Usage of Predis library
- [Predis](https://github.com/nrk/predis) is the most powerful and popular client library for Redis written in PHP. QCubed uses Predis for using Redis as cache. QCubed's cache system cannot use Redis unless you install Predis. It is also a suggested package in our `composer.json`.
+ [Predis](https://github.com/nrk/predis) is a popular (and widely regarded one of the most powerful) library for Redis written in PHP. QCubed uses Predis for using Redis as cache. QCubed's cache system cannot use Redis unless you install Predis. It is also a suggested package in our `composer.json`.
  
  You can simply use `composer require "predis/predis":"^1"` to install. Please see the [project page](https://github.com/nrk/predis) for installation instructions.
  
--- a/docs/Caching/Redis.md
+++ b/docs/Caching/Redis.md
@@ -26,8 +26,69 @@ define ('CACHE_PROVIDER_OPTIONS', serialize(
 
 The configuration **must always contain the 'parameters' and 'options' keys**. These go in the constructor of `Predis\Client`. Please see the [project page](https://github.com/nrk/predis) if you want to use custom connection options for your redis-server installation.
 
+### Expiration of keys
+Keys can be set to auto-expire in Redis (like in other cache service providers). However, there is a special behavior for Redis cache controlled by a named constant - `_REDIS_CACHE_PROVIDER_DEFAULT_TTL_`. We will first tell you why we treat Redis cache differently.
+
+Of all cache providers, Redis is the only one which **persists its data on disk**. *LocalMemory* provider would invalidate caches after request ends. *NoCache* invalidates caches immediately. *APC* would invalidate after webserver/fpm service restart. *Memcache* would invalidate after memcache service or server restarts. 
+
+By default, Redis does not exhibit such a behavior. All keys will go on getting saved until they are manually deleted or are evicted because of a expire time applied to the key when setting the value. Since caches should be automatically managed and should use limited resources, it is important that they free resources from time to time. Each cache provider has these limitations in place while a server-reboot is the final solution to a error or resource-overusage for them. The only method Redis' default installation gives us for automatically managing resource usage (in this case, storage) is setting an expire-time on keys using which Redis automatically frees up memory/storage. The `_REDIS_CACHE_PROVIDER_DEFAULT_TTL_` named constant allows us to control Redis' storage requirement. 
+
+When this value is set, each cache key set into Redis by using QCubed's built in Cache-Provider mechanism gets a default expiration time. The expiration time is equal to the number of seconds this value is defined as. 
+  
+**Example**: 
+
+You have defined the value in `configuration.inc.php` as:
+
+```php
+define('_REDIS_CACHE_PROVIDER_DEFAULT_TTL_', 7200);
+```
+
+Later, when you save a value into Redis cache using:
+
+```php
+QApplication::$objCacheProvider->Set('someKey1', $strSomeValue, 60);
+```
+
+Then *someKey1* will automatically be removed from cache in **60 seconds**. This is because you passed the third argument into the method as 60, which overrides the defaults.
+
+However, if you were to set another key using: 
+
+```php
+QApplication::$objCacheProvider->Set('someKey2', $strSomeValue);
+```
+
+Then *someKey2* will be removed from the cache in 2 hours (7200 seconds).
+
+---
+If you set the value to 0 (or anything else which is evaluated as 0 by PHP), like: 
+
+```php
+define('_REDIS_CACHE_PROVIDER_DEFAULT_TTL_', 7200);
+```
+
+This this statement: 
+```php
+QApplication::$objCacheProvider->Set('someKey3', $strSomeValue);
+```
+
+Will not set a default expiration time and the value corresponding to *someKey3* will stay in Redis cache untill you remove it.
+
+However, this statement: 
+
+```php
+QApplication::$objCacheProvider->Set('someKey4', $strSomeValue, 60);
+```
+
+would still make *someKey4* expire after 1 minute.
+
+
 ## Full power of Redis is available
  
- The `QApplication::$objCacheProvider` created for Redis Caching is a wrapper around `Predis\Client` instance and can take all commands like `sadd` and `sismember` (e.g. `QApplication::$objCacheProvider->sadd('keyS', [2,5,10])`). 
+ The `QApplication::$objCacheProvider` created for Redis Caching is a wrapper around `Predis\Client` instance and can take all commands like `sadd` and `sismember` For example:
+  
+```php
+  QApplication::$objCacheProvider->sadd('keyS', [2,5,10]) 
+```
+would save 2,5 and 10 into a *Set data structure* named *keyS*.
  
  If you have a persistent Redis instance, you can use the caching system to use the entire Redis functionality using the Predis library.
--- a/docs/Caching/Redis.md
+++ b/docs/Caching/Redis.md
@@ -1,0 +1,33 @@
+# Redis Caching
+
+Redis is a key-value style data-structures store. The key length as well as the value can be as large as 512 MB. The 'value' in Redis can be one of the multiple [data structures](https://redis.io/topics/data-types). 
+ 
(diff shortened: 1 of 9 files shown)


spekary:
If we are going to make this part of 3.1, you will need to make the pull request against the 3.1 branch.

docs/Caching/Redis.md
```diff
@@ -0,0 +1,33 @@
+# Redis Caching
+
+Redis is a key-value style data-structures store. The key length as well as the value can be as large as 512 MB. The 'value' in Redis can be one of the multiple [data structures](https://redis.io/topics/data-types). 
+ 
+ Performance of Redis depends on whether it is running in persistent mode (saving data to disk) or in-memory mode (all contents are within the specified memory limits). Default configurations run in persistent mode. However, depending on use-case (data structures stored, amount of data stored etc.), performance of Redis can be between "a little slower" to "just a little faster" than memcached. However, implementation of Redis is way more powerful because of the data-structures implementation.
+ 
+ ## Usage of Predis library
+ [Predis](https://github.com/nrk/predis) is the most powerful and popular client library for Redis written in PHP. QCubed uses Predis for using Redis as cache. QCubed's cache system cannot use Redis unless you install Predis. It is also a suggested package in our `composer.json`.
```
spekary: You shouldn't put opinions in your readmes. Whether its true or not, this might change in the future. Just state the facts.

docs/Caching/Redis.md
```diff
@@ -0,0 +1,33 @@
...
+			'options' => array()
+		)
+	));
+```
+
+The configuration **must always contain the 'parameters' and 'options' keys**. These go in the constructor of `Predis\Client`. Please see the [project page](https://github.com/nrk/predis) if you want to use custom connection options for your redis-server installation.
+
+## Full power of Redis is available
+ 
+ The `QApplication::$objCacheProvider` created for Redis Caching is a wrapper around `Predis\Client` instance and can take all commands like `sadd` and `sismember` (e.g. `QApplication::$objCacheProvider->sadd('keyS', [2,5,10])`).
```
spekary: Think about other ways to use the cache too. $objCacheProvider sepecifically caches certain database calls. It should not be required to set $objCacheProvider to use Redis. QWatcher needs a key-value store as well, and this would work well for that too. This is mostly something to inform your readers of, and would not likely effect your code.

includes/framework/QCacheProviderRedis.class.php
```diff
@@ -0,0 +1,96 @@
...
+		/**
+		 * Set the object into the cache with the given key
+		 *
+		 * @param string $strKey                the key to use for the object
+		 * @param string $objValue              the object to put in the cache
+		 * @param null   $intExpireAfterSeconds Number of seconds after which the key will be expired
+		 *
+		 * @return void
+		 */
+		public function Set($strKey, $objValue, $intExpireAfterSeconds = null) {
```
spekary: See QCacheProviderAPC. We have a global variable ttl that manages the expiration time. Do you see a value in needing to manage the expiration time on every single Set, or will a global variable work so that the developer doesn't have to set this every time?

includes/framework/QCacheProviderRedis.class.php
```diff
@@ -0,0 +1,96 @@
...
+		/**
+		 * Let other redis methods work
+		 *
+		 * @param $commandID
+		 * @param $arguments
+		 *
+		 * @return mixed
+		 */
+		public function __call($commandID, $arguments) {
+			return $this->objPredisClient->executeCommand(
```
spekary: Nice.

spekary:
If we are going to include this in unit tests, we will need a way to detect whether the dependencies are installed in the system before attempting to run the test. We want to be able to run our unit tests on any developers local system, and on travis, without requiring the local developer to install unneeded dependencies. I am assuming travis can install redis.

docs/Caching/Redis.md
vaibhav-kaushal: Which statement do you think is my personal opinion? The one about "most powerful library" (I guess)?

docs/Caching/Redis.md
vaibhav-kaushal: I have more ideas in mind. Like using it for making a set of methods which work on data structures (sets, lists, sorted sets, HyperLogLogs), Using Redis as FormStateHandler etc. When that happens, we can add those in the codebase and update the documentation around it. Since it is not available now, there is no reason we should not mentioning this in the current documentation.

includes/framework/QCacheProviderRedis.class.php
vaibhav-kaushal: For Redis, I have updated the config file and put in the `_REDIS_CACHE_PROVIDER_DEFAULT_TTL_` named variable which works for only the cache (you can see the usage in this file as well). This is useful for caches. 

As far as the usability of **per set** is concerned, I see (in my application, at least) that sometimes we need to create short-living objects (like a list of objects which should be cached and updated every 10 minutes) while there are some which need updating a lot less frequently. In such a case, one may want to fine-tune the duration for which the value has to be cached.

Think like this: There is a post which we present to the client(s). That post is seldom going to change - so that post should be cached for as long as possible (a day's TTL would be fine for that one). At the same time, we calculate *6 other posts related to this post*. We want this list to change every 5 minutes so that the user sees new recommendations every 5 minutes. This thing should have a shorter TTL. This parameter enables such behavior. 

This is the reason I added the option.

docs/Caching/Redis.md
vaibhav-kaushal: Updated it in the latest commit.

vaibhav-kaushal:
@spekary I think yes, Travis can install Redis. I am sure it is working - using it in one of my production environments. But this is a framework so tests are important. 

If we are to use this one in our tests, we will also have to move `predis/predis` from `suggest` group to `require` group.

vaibhav-kaushal:
@spekary , @olegabr (and others) Please review this PR again and merge when possible.

olegabr:
> If we are to use this one in our tests, we will also have to move predis/predis from suggest group to require group.

This makes me think that functionality like this PR should be provided as plugins. What do you think?
@spekary, do you have some suggestions of how to implement it as a plugin?

vaibhav-kaushal:
@olegabr Going by that logic, we would have to use **MySQL** and **PostgreSQL** implementations as plugins too. But they are used as 'adapters'. In today's backend systems, caching is paramount. Actually, even for a mid-sized business, caching and basic analytics are needed. 

I am in favor of making this one a hard requirement. I would like to explain more on this one:
----
A framework or library is supposed to be a collection of methods grouped in a certain fashion to facilitate a bunch of related functionality which can fit a number of use-cases. 

In our case, when we say "we support databases", a natural question arises: which databases. We say *MySQL* and *PostgreSQL*. Then another question that arises for the reader is: What do I have to do to use them? We answer that with **nothing**. As a user, I like that **nothing**. It gives me the confidence that the functionality is well-supported and integrated into what I am downloading and that it's not something that is going to break because it was another plugin lying around which no one cared about.

When we say "we support caching", the same (and natural) question arises: which caching systems? The answer has to be finite set. We have **Local Memory Cache** and **Memcache** right now in the core. If you are to give that confidence to the user that Caching system would work well, the system should be integrated into the framework by default. The only difference **Predis** has from **Memcache** is that memcache is a *PHP extension* while Predis is a *third party library*.

As a user I do not care how the thing gets installed. I care that it works without flaws. Abstracting the implementation details is at the heart of a framework. I am in favor that we make Predis a hard requirement so that other things which we can build on top of Redis could also be done with confidence. 

The following things are planned (on my end) for being built atop Redis: 

1. FormStateHandler with Redis as the storage backend: Since Redis can act as a data store, using it as a FormStateHandler would be good. It is also fast, easy to install and requires minimal attention after it starts running. Since Redis can both persist data and auto-evict it based on keys and a TTL, Redis relieves the programmer from worrying about Garbage Collection of FormStates, timing, performance impact etc. Also, per formstate can be as large as 512 MB - and that's more than enough!
2. One-to-many Relationships scenario: Assume you want to associate "likes" on a "post". This is a one-to-many relationship scenario where "ordering" is mostly not required. Redis is great at doing this with its **sets** and **lists** data structures. This one is low-priority and still not well thought-about.
3. Data Structures: Sometimes we need to implement data structures for objects in the application and the built-in functions in PHP are not enough. Stacks, Queues, Double-Ended-Queues, Lists, Sets, Graphs, Trees etc. Many of these are already supported by Redis. Some are not but Redis' features allow these to be done more easily. We already have `QStack` class and `QTreeNav` control. Allowing DataStructures operation on objects is something which really enables the developers to move fast. Redis helps here tremendously. 

For these cases, I would recommend this to remain in the core.

includes/framework/QCacheProviderRedis.class.php
spekary: OK. Document if possible how to set up the ttl default value, and that there is a default, so developers know what will happen when they don't specify one.

spekary:
A couple of thoughts:
- I don't mind this being in an adapter, so that its always installed just like the other cache adapters. Its not very much code at this point. Advanced features that are Redis specific could be put in a plugin without much trouble. 
- I DO mind having the redis-php github repo be required.  

> If we are to use this one in our tests, we will also have to move predis/predis from suggest group to require group.

You actually can install it without moving it to the require group. In travis.yml, in the install section, put the composer require command for the php redis repo on the line after "- composer install". That way we can include it in our tests without requiring everyone to install it. The tests themselves should check for the presence of that code before running the test so that the basic tests can be run by any developer regardless of having redis. You could output a message "Redis not installed. Unit tests skipped."

So, to merge this, please do the following: 
- doctor up travis.yml as I described
- add the unit tests
- if possible, structure the unit tests so that they can be used on other cache providers, like the APCu provider.

We should be good to go then.

vaibhav-kaushal:
@spekary Done. Please review again and merge if you think this fits.

spekary:
Reviewing the code again, everything looks good, except there is another major issue. You are introducing a major change to the project, which is the "docs" folder, yet another place for documentation, that is not attached to the actual code it documents. You then make the statement that this is where all the documentation is going to go.

You should try to avoid combining multiple changes like this. One will block the other. In this case, checking in the Redis code is blocked by needing a discussion on changing our documentation practices.

I understand your frustration with our current documentation strategy. We currently either document through the examples page, which isn't a good place for completely documenting a feature, or in the code itself, which is hard to extract at this point and also hard to format. There is a little bit of effort to document things in the Wiki, and since that is a common practice for other github projects, that is probably what we should do too.

Since you wrote this up in markdown, my inclination is that this should go into the Wiki, and you can put it a pointer to it from the code. It also might be good to have an example page, but you can add it later.

Perhaps its time to have a significant discussion on documentation elsewhere, but for now, can you remove the statements that imply that this is where all future documentation is going to go? No other frameworks document their code this way.

vaibhav-kaushal:
@spekary I understand your point. For the discussion over Documentation (and its reasoning), I have created #1154. I have edited the main file where I think there was an implication.

PR #1156, opened 2017-01-23
feature-BlockingUnwantedEvents

spekary:
This feature provides a mechanism to "debounce" events. It solves the classic problem of the "submit" button that you only want the customer to press once. Its also useful for events that bring up a dialog, so that the dialog is only displayed once.

Additionally, this updates the ajaxq ajax event queuer to code that is being currently maintained by the developer and that uses the more modern "promises" to manage the queue. It also has the advantage that it works without modification for our event queue system. (we had to modify the old ajaxq code to work for us).


kukrik:
When merged? Where would you put a version of what kind? Whether to version 3.0 or not?

spekary:
Not 3.1 since its a change to the event mechanism I think.

spekary:
I have merged current 3.0 into 3.1, so if anyone wants to test 3.1, we could release it and then bump newer changes to 3.1, though we have one issue to fix in the examples.

kukrik:
If you can, it would be very nice to see it in version 3.0.

PR #1158, opened 2017-01-26
Legacy documentation __toString removed

matthiaz:
I wouldn't tell the user to use __toString as a function call, I don't think it's a good idea to be calling it like that.


vaibhav-kaushal:
I agree. Casting to string would automatically invoke it anyway!

Issue #1160, opened 2017-01-30
Move Base64Encode method from QCryptography to QString


vaibhav-kaushal:
This method is a very useful one. Also, it actually operates on strings and is not really a cryptographic method per se. e.g.
```
$strBase64Encoded = QCryptography::Base64Encode($strSomeString);
```
does not make sense!

Moving this method to QString would be fruitful.

matthiaz:
Maybe a stupid question. But why would you need a base64 function. Isn't
base64_encode(), the native van php function good enough?

Op 30 jan. 2017 12:14 p.m. schreef "Vaibhav Kaushal" <
notifications@github.com>:

vaibhav-kaushal:
Here is the method: 

```php
protected static function Base64Encode($s) {
		$s = base64_encode($s);
		$s = str_replace('+', '-', $s);
		$s = str_replace('/', '_', $s);
		$s = str_replace('=', '', $s);
		return ($s);
	}
```

It replaces `+` and `/` to `-` and `_` respectively. The replaced characters are safe for being sent in the URL without causing unwanted effects. That's why this function is better. No other reason whatsoever.

matthiaz:
why not use urlencode?

2017-01-30 13:57 GMT+01:00 Vaibhav Kaushal <notifications@github.com>:

spekary:
Url encode works just for URLs. This function has the specific purpose of preparing a binary object for use as a PUT variable on a web page, if you render the string out into a hidden input. Since base64 has a limited character set, the characters listed in the function are the only ones that need to change, so this is as efficient as possible. This function is used by the default form state handler so that the form state is passed from submit to submit in a hidden input. Since that is a really bad way to manage the form state, this isn't a high priority function. It certainly could be moved to the QString utility, or the QHTML class, but really, this is no big deal. Perhaps just comment it a little more and leave it?

matthiaz:
I'm voting for commenting and leaving it alone for now.

2017-01-30 19:03 GMT+01:00 Shannon <notifications@github.com>:

vaibhav-kaushal:
@matthiaz I can understand your reservations around it. I am not sure if you are using it. Are you? Please let me know. This method as such has not been used elsewhere within the framework. 

@spekary I would like to place this method in a better place. I think calling it `QString::UrlSafeBase64Encode` is okay (though a little awkward). Suggest something better if you have a better name in mind. Will create a PR soon.

vaibhav-kaushal:
[...]

PR #1161, opened 2017-01-31
Refactor Base64Encode and Base64Decode methods in QCryptography

vaibhav-kaushal:
Changes:
 - Move the methods to QString class.
 - Documentation 

Solves #1160

### DIFF
--- a/docs/Strings/README.md
+++ b/docs/Strings/README.md
@@ -1,0 +1,47 @@
+# Strings Treatment in QCubed
+
+*This direcotry contains documents describing methods used for working with **string** in QCubed.*
+ 
+ ## QString Class
+ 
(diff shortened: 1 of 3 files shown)


vaibhav-kaushal:
Will merge this in case no one points out problem with it.

spekary:
Please hold off. I use those routines, so I need to make sure. Just in the middle of a deadline currently.

spekary:
Also, this is a breaking change, so should be merged into 3.1. Or, you can put stubs in QCryptography that point to the new code and deprecate those stubs in the comment so that anyone using those routines directly will not break. In that case, it can be merged into 3.0.

vaibhav-kaushal:
@spekary Stubs makes most sense. Let me update the PR.

vaibhav-kaushal:
@spekary Done.

vaibhav-kaushal:
It has been a week and all the requirements have been met. @spekary Awaiting your confirmation.

PR #1163, opened 2017-02-06
bugfix-DebounceTableRows

spekary:
Fixing List generators to prevent a problem when clicking on rows that are supposed to create actions.


spekary:
Failed build is due to problem that has been fixed in the v3.1 branch. The merged version should pass unit tests fine.

PR #1166, opened 2017-02-08
Redis Formstate Handler completed

vaibhav-kaushal:
Redis Formstate Handler completed. Tested and works for me.

@matthiaz You were interested in this feature. Can you please test and tell us if it works well enough?

### DIFF
--- a/includes/qcubed.inc.php
+++ b/includes/qcubed.inc.php
@@ -80,6 +80,7 @@
 	QApplicationBase::$ClassFile['qsessionformstatehandler'] = __QCUBED_CORE__ . '/qform_state_handlers/QSessionFormStateHandler.class.php';
 	QApplicationBase::$ClassFile['qfileformstatehandler'] = __QCUBED_CORE__ . '/qform_state_handlers/QFileFormStateHandler.class.php';
 	QApplicationBase::$ClassFile['qdbbackedformstatehandler'] = __QCUBED_CORE__ . '/qform_state_handlers/QDbBackedFormStateHandler.class.php';
+	QApplicationBase::$ClassFile['qredisbackedformstatehandler'] = __QCUBED_CORE__ . '/qform_state_handlers/QRedisBackedFormStateHandler.class.php';
 
 	// Define ClassFile Locations for Framework Classes
 	QApplicationBase::$ClassFile['qrssfeed'] = __QCUBED_CORE__ . '/framework/QRssFeed.class.php';
--- a/includes/qform_state_handlers/QRedisBackedFormStateHandler.class.php
+++ b/includes/qform_state_handlers/QRedisBackedFormStateHandler.class.php
@@ -1,0 +1,180 @@
+<?php
+
+	/**
+	 * This will store the formstate in a Redis DB.
+	 *
+	 * This offers significant speed advantage over PHP SESSION because EACH form state
+	 * is saved using a unique key with Redis, and only the form state that is needed for loading will
+	 * be accessed (as opposed to with session, ALL the form states are loaded into memory
+	 * every time). This is very similar to how DbBackedFormStateHandler works, only this one uses
+	 * Redis for storing the FormStates
+	 *
+	 * The downside is that because it doesn't utilize PHP's session management subsystem,
+	 * this class must take care of its own garbage collection/deleting of old/outdated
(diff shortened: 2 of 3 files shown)


kukrik:
@vaibhav-kaushal, It would be nice to see one example or accurate manual?

This is certainly very useful for me, if the database grows to big...

vaibhav-kaushal:
@kukrik Just install Redis on your box and change the formstate handler in config file. It should work. :)

kukrik:
Do I understand that, for example, Draft level also works faster when the list will be very long? I thought it must be a bit of code to change the level, etc...

If so, then it would be very awesome?!

vaibhav-kaushal:
Try it yourself. I think it will be a little faster compared to DbBackedFormstateHandler. I have not done performance test yet.

kukrik:
I was tested it. It seems to be a bit fast. Better is, if @matthiaz is testing its own method. We will hear from him.

matthiaz:
I tested the Redis formstate handler. A couple of observations. 

How to install? 
`yum install redis` (or apt-get install redis on debian systems)
`service start redis`

Then I had to add 
`"predis/predis": "dev-master"`
to composer.json 
and run
`composer update`

Do we mention how to install/configurate it in the documentation?

Idea: maybe we could clarify the error on line 46 with information on how to install Predis?
_Predis library needs to be installed for Redis Formstate Handler to work._ 


When I switch to the Redishandler I immediately get **gzuncompress(): data error** 
Line: 167 : $strSerializedForm = gzuncompress($strSerializedForm);
So $strSerializedForm must somehow, not be a gz? I having trouble debugging this. Any ideas?

I tried restarting redis: service redis restart (which should clear up any data, right?)
But the same thing happens over and over again. When switching back to SessionStateHandler it works.

kukrik:
@matthiaz, I do not know how you are using the developer's server. I use in order Homebrew.

I was easy to use, for example, the command line to `brew install redis` and voilaa.

After the command line sudo add composer `sudo composer require predis/predis`.

How are you?

matthiaz:
I'm using Linux, not mac ;-)

some thing though: yum is the package manager I use in stead of Homebrew.
No problem there.

composer require simply adds data to composer.json, so that does the same
as well.

2017-02-09 15:41 GMT+01:00 Tiit Papp <notifications@github.com>:

kukrik:
@matthiaz, I see it now :)! I watched part of your code (`"predis/predis": "dev-master"`). It seems to be different. Since I not write directly to composer.json. In some cases though I write... Then mine looks like: `"predis / predis" "^" 1.1`.

Try it!

vaibhav-kaushal:
@matthiaz No, restarting Redis will **not** clear anything. Redis is a **persistent** data store. You will have to install `redis-cli` program and then run `redis-cli flushdb` from the command line (given you have not customized Redis server's config ).

Secondly, I have tested Predis library using version 1.1.1 and hence I locked the version to "^1.1". 

Thirdly, I am not sure if we should tell the user how to install Predis or not because installation instructions can change and more importantly, one can also search for "predis" on google and he will get to know how to install Predis. One can always raise the issue here and we can guide, though installation of Predis is quite simple - using composer which one would use to install QCubed anyway.

As far as the gzuncompress is concerned, I would recommend you to switch to the Redis Formstate Handler and then open a fresh browser window and start browsing the project again. Let me know if that works (or not).

matthiaz:
Oh, I didn't know redis was persistent. I thought it wasn't because it was
a "memory" storage.
Anyway, I ran redis-cli flushdb -> returned OK. I didn't do any
customisations on the configuration.

I tried in different browser windows and even different browser (chrome and
firefox, same thing :( )


2017-02-09 16:00 GMT+01:00 Vaibhav Kaushal <notifications@github.com>:

vaibhav-kaushal:
@matthiaz Did you retry with gzuncompress removed? I have not encountered this problem yet - I am not able to reproduce it.

matthiaz:
I'm away from my computer for the next week, I'll check again when I
return.

Op 10 feb. 2017 3:24 p.m. schreef "Vaibhav Kaushal" <
notifications@github.com>:

spekary:
I am going to commit this since its something you can optionally turn on and not required for QCubed, and we can fix individual problem later.

PR #1167, opened 2017-02-09
Generate random string of a given length from a character set

vaibhav-kaushal:
GetRandomString method in QString is new and facilitates the feature

It returns a string of a given length, the characters of which are extracted
from the character set supplied as the second argument and are in random
order. In case a character set is not supplied, alphanumeric character set
is used by default.

### DIFF
--- a/includes/framework/QString.class.php
+++ b/includes/framework/QString.class.php
@@ -237,20 +237,10 @@
 			$intLength = QType::Cast($intLength, QType::Integer);
 			$strCharacterSet = QType::Cast($strCharacterSet, QType::String);
 
-			// Get Characters into array
-			$strCharacterArray = str_split($strCharacterSet);
-			// Get Unique characters
-			$strCharacterArray = array_unique($strCharacterArray);
-
-			// String to return
-			$strToReturn = '';
-
-			// Build the string
-			$intCharacterCount = count($strCharacterArray);
-			for($i=0; $i<$intLength; $i++) {
-				$strToReturn .= $strCharacterSet[rand(0, ($intCharacterCount-1))];
-			}
-
-			return $strToReturn;
+			return substr(
+				str_shuffle(
+					str_repeat($strCharacterSet, ceil($intLength / strlen($strCharacterSet)))
+				)
+				, 0, $intLength);
 		}
     }
--- a/includes/framework/QString.class.php
+++ b/includes/framework/QString.class.php
@@ -223,4 +223,34 @@
             $s = base64_decode($s);
             return ($s);
         }
+
+		/**
+		 * Get a random string of a given length
+		 *
+		 * @param int $intLength       Length of the string which is to be produced
+		 * @param int $strCharacterSet Character Set to be used
+		 *
+		 * @return string The generated Random string
+		 */
+		public static function GetRandomString($intLength, $strCharacterSet = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789') {
+			// Cast in case there were something else
+			$intLength = QType::Cast($intLength, QType::Integer);
+			$strCharacterSet = QType::Cast($strCharacterSet, QType::String);
+
+			// Get Characters into array
+			$strCharacterArray = str_split($strCharacterSet);
+			// Get Unique characters
+			$strCharacterArray = array_unique($strCharacterArray);
+
+			// String to return
+			$strToReturn = '';
+
+			// Build the string
+			$intCharacterCount = count($strCharacterArray);
+			for($i=0; $i<$intLength; $i++) {
+				$strToReturn .= $strCharacterSet[rand(0, ($intCharacterCount-1))];
+			}
+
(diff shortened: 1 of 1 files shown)


matthiaz:
Hi @vaibhav-kaushal ,

Nice addition. I have a similar function in our own code base, I'd like to share. It's a little different but I think it does the same thing. I tried to avoid the loop and repeated calls to rand. I don't know if it is faster/slower though. Do we have a performance testing thingy somewhere to measure stuff like this?

```
public static function RandomString($length=6,$chars='abcdefghijkmnpqrstuvwxyz23456789ABCDEFGHJKMNPQRSTUVW;,.:<>!@*()+=_-|{}[]') {
	return substr(
		str_shuffle(
			str_repeat($chars, $length)
		)
		,0,$length);
}
```

Let me know what you think.

vaibhav-kaushal:
I think this is brilliant. I did not remember about str_shuffle method, at all (there are so many of those). However I made a modification to your program as such: 

```php
public static function RandomString($length=6,$chars='abcdefghijkmnpqrstuvwxyz23456789ABCDEFGHJKMNPQRSTUVW;,.:<>!@*()+=_-|{}[]') {
	return substr(
		str_shuffle(
			str_repeat($chars, ceil($length / strlen($chars))
		)
		,0,$length);
}
```
This makes sure that we are making the `str_repeat` method work only as much as is required.

Here are the results for generating **a million** Random strings, each of **100 characters** using three methods:

```
Time taken by       MY METHOD for  1000000 generations = 57.893663167953 seconds
Time taken by MATTHIAZ METHOD for  1000000 generations = 3.1451518535614 seconds
Time taken by OLD MATT METHOD for  1000000 generations = 54.131749153137 seconds
```
Here 
 - **MY METHOD** is the algorithm I have submitted in this PR. 
 - **MATTHIAZ METHOD** is the updated algorithm I have presented in this very post
 - **OLD MATT METHOD** is the algorithm you originally submitted

I think your and mine both algorithms were slow (mine being tad slower) but updating yours with the ceil function makes it significantly faster, especially as the length of the string goes up! 

Below is the script used for testing. Run it by your own to see the performance differences. 

--------
```php
<?php
	include 'qcubed.inc.php';

	function GetRandomString($intLength, $strCharacterSet = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789') {
		// Cast in case there were something else
		$intLength = QType::Cast($intLength, QType::Integer);
		$strCharacterSet = QType::Cast($strCharacterSet, QType::String);

		// Get Characters into array
		$strCharacterArray = str_split($strCharacterSet);
		// Get Unique characters
		$strCharacterArray = array_unique($strCharacterArray);

		// String to return
		$strToReturn = '';

		// Build the string
		$intCharacterCount = count($strCharacterArray);
		for ($i = 0; $i < $intLength; $i++) {
			$strToReturn .= $strCharacterSet[rand(0, ($intCharacterCount - 1))];
		}

		return $strToReturn;
	}

	function RandomString($length = 6, $chars = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789') {
		return substr(
			str_shuffle(
				str_repeat($chars, ceil($length / strlen($chars)))
			)
			, 0, $length);
	}

	function RandomStringOLD($length = 6, $chars = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789') {
		return substr(
			str_shuffle(
				str_repeat($chars, $length)
			)
			, 0, $length);
	}

	$num = 1000000;
	$intLengthToGenerate = 100;

	$strStart = microtime(true);
	for ($i = 0; $i < $num; $i++) {
		GetRandomString($intLengthToGenerate);
	}
	$strEnd = microtime(true);
	echo 'Time taken by       MY METHOD for  ' . $num . ' generations = ' . ($strEnd - $strStart) . ' seconds' . PHP_EOL;

	$strStart = microtime(true);
	for ($i = 0; $i < $num; $i++) {
		RandomString($intLengthToGenerate);
	}
	$strEnd = microtime(true);
	echo 'Time taken by MATTHIAZ METHOD for  ' . $num . ' generations = ' . ($strEnd - $strStart) . ' seconds' . PHP_EOL;

	$strStart = microtime(true);
	for ($i = 0; $i < $num; $i++) {
		RandomStringOLD($intLengthToGenerate);
	}
	$strEnd = microtime(true);
	echo 'Time taken by OLD MATT METHOD for  ' . $num . ' generations = ' . ($strEnd - $strStart) . ' seconds' . PHP_EOL;

```

vaibhav-kaushal:
I am going to use your method with the `ceil` function thrown in! It becomes very fast that way!

matthiaz:
Nice!

Do you mind if I steal your ceil() addition and implement it in my own code?

My company could use a PHP/qcubed a developer like you if you're interrested ;-)

vaibhav-kaushal:
@matthiaz Sure. It just went public anyway ;)

PR #1169, opened 2017-02-10
feature-30compatibility

spekary:
In preparation for 3.1 release, this adds backwards compatibility with 3.0 projects and adds a documented path to migrating code to 3.1 for the privateColumnVars feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,52 @@ Through its plugin system, QCubed makes it easy to package and deliver enhanceme
 
 The installation procedure is described in detail here: [Installation instructions](https://github.com/qcubed/qcubed/blob/master/INSTALL.md "Installation instructions").
 
-* * *
+## Upgrading
+###2.x -> 3.0
+3.0 was a major architectural change from 2.x. You should essentially start over by creating a new project, 
+generating your models, using the ModelConnectorEditor to refine what is generated in the connectors(used to be called MetaControls),
+and then copying code from your old version to the new version. You will find that many of the things you had to do
+by hand are now done in generated code, so it might not take as long as you think. Lets hop :-)
+
+###3.0 -> 3.1
+To help with the transition, the new private variable feature is turned off by default. You will
+need to turn it on by editing the codegen_settings.xml file and adding a 'privateColumnVars="true"' parameter to the createOptions tag.
+See the codegen_settings.xml file in the qcubed/install/project/configuration directory for an example.
+
+This will make all of the protected column variables that are in the Gen class private, so you will not be able to access them
+directly from your Model subclasses. For example, if you have a "Name" column in a table, you can do:
+ ```php
+ $strName = $this->Name;
+ ```
+ or better yet:
+```php
+$strName = $this->getName();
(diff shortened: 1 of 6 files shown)


vaibhav-kaushal:
Why are we trying to make all member variables private!? Isn't `protected` good enough? We get to keep the performance high with the subclass ORM file while making sure external modifications are not possible?

spekary:
I write about it in a couple of places, but the new ORM has a few goals:
1) When creating an UPDATE clause, only update the fields of a record that have changed. In 3.0-, the entire record is saved every time an update happens.
2) There is a kind of bug in 3.0- having to do with the QSelect clause. You can QSelect just a few fields, but you can later refer to a field that was not selected, and get a default value, not knowing that the data did not come from the database. I want to throw an exception if you accidentally refer to data that is not loaded.
3) The Optimistic Locking exception will get thrown even if you are modifying a portion of a record that did not conflict with a previous edit, because we currently update the entire record. By updating only what is changed, we might be able to avoid a conflict with a previous edit.

For all of these to work, the ORM must keep track of what fields in a loaded model record are valid or have changed. To be able to do this, it must make the member variable corresponding to the field private, or it will be too easy for a develper to accidentally circumvent the processes and get really bizarre behavior. For example, you could modify a variable, and then when you update it, it doesn't get updated because a change is not detected.

Regarding performance, that should not be a problem. Using the new getters and setters, access to the variables is very fast. PHP7 caches the function call, and I believe HHVM may even inline them. Regardless of any minor speed changes though, the data security and prevention of unexpected behavior is a big win I think.

vaibhav-kaushal:
@spekary I think the goals are set in the right direction. 

**Related, yet offtopic**: I have an added goal for the ORM Level:

Generation of methods around atomic operations on primitive data types. This includes increments and decrements. Incrementing and Decrementing a numeric value (normally integers) by a certain value (normally 1) is one of the most common operations on those types. The best example would be keeping a 'hit count' on a post or implementing a count-down (either timely or resource based).

Right now, we do it by loading the object, setting the field to field-1 in the program logic and then running save on the object. While selective updates that you want to work on in 3.1 solve this problem to a great extent, I think having methods for these operations on numeric fields would be great. 

That being said, I realize that we would have to give the developer the option of "what features do you want in your codebase" before we actually run the codegen because not everyone is going to want the incrementation feature, and of course not on all integer fields there are in all tables!

vaibhav-kaushal:
I think you can merge this. Looks harmless to me. I am on custom templates right now so cannot verify straight-away. May be @olegabr can help in this?

spekary:
re: hit counting. Feel free to do something along that lines, but in the end, it may need to change. I have something in the works around using the QQuery language for updating, but it will take some time to implement.

Also, I am not sure how global a need it is. Writing a SQL function "Update T set a= a+1 where b=X" really is not that difficult to do manually. Putting a lot of work in to replacing something like that doesn't seem like its worth the effort. We certainly have more pressing things to do, like documentation and namespacing.

PR #1173, opened 2017-02-11
feature-HtmlTable

spekary:
Feature to quickly create an html table from an array of data.

### DIFF
--- a/includes/framework/QHtml.class.php
+++ b/includes/framework/QHtml.class.php
@@ -450,4 +450,63 @@
 			return nl2br(htmlspecialchars($strText, ENT_COMPAT | ENT_HTML5, QApplication::$EncodingType));
 		}
 
+		/**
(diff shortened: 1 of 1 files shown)


vaibhav-kaushal:
How do I use it? Can you show an example (in the conversation)?

spekary:
I put a usage example in the comments and improved the comments a bit. This is a utility function, so there are many usage scenarios, but it is just solving the need to output an array of arrays, or an array of objects as a quick HTML table without needing to create a QHtmlTable object and add columns and a data binder.

The most basic example would be:

```
$data = [['apple', 'fruit'], ['carrot', 'vegetable']];
print (QHtml::RenderTable ($data));
```
It is flexible enough to be able render tables with key=>value pairs, or objects with $obj->Value values. See the description of the function parameters.

PR #1174, opened 2017-02-11
Changed delimiter for Iv and Hashkey

vaibhav-kaushal:
The '=' character is padding character in raw base64 encoded string.
It is removed in the QString::Base64UrlSafeEncode method (or the same implementation
in QCryptography class earlier). When trying to use base64 encoding built into
QCryptography class, the decryption fails because the base64 encoded data will
eventually lack the delimiting character.

The ':' character is outside the base64 character set and thus, when encoding,
the delimiter would not be lost.

I am facing that as a major issue with `=`. Changing it to `:` makes it work.


vaibhav-kaushal:
@spekary PLease look into this once. It has been working well in my tests while the `=` version was causing me troubles.

PR #1186, opened 2017-02-28
bugfix-OptimisticLockingIssues

spekary:
For too long has QCubed issued optimistic locking errors for back button presses on the browser, or multiple people editing a record.

This fix does the following:
- Puts dialogs in to the codegenned forms and dialogs to correctly detect optimistic locking errors and asks users what to do when a record has changed that the user is editing.
- Prevents the Back button on the browser from causing an Optimistic Locking error by making sure changes are posted to the form before proceeding to the next page. 
- Fixes examples page to work
- Fixes the default QWatcher to work on initial installation, which is needed for forms that use a dialog to do edits.
- Wraps Update code in transactions in situations where an update requires changes to multiple tables. This prevents data corruption and missing optimistic locks when multiple users are updating the same record.


spekary:
Travis is currently broken. The problem is an AWS outage. Don't merge this yet.

spekary:
Travis appears to be fixed now.

PR #1191, opened 2017-03-05
doc-initial

spekary:
Adding initial documentation at a project level.


vaibhav-kaushal:
Why don't we also merge it into the current master?

spekary:
This will become the master soon enough. I want to make sure the docs correctly reflect the state of the examples, and the rest of the website.

PR #1210, opened 2017-03-15
bugfix-NullTimestamp

spekary:
Allowing timestamps to be set to null, even when not allowed to be null, since MySQL allows this to update the timestamp.


spekary:
Creates a problem with optimistic locking checks. Working on it.

PR #1221, opened 2017-03-20
Cleaning up the QDbBackedSessionHandler

matthiaz:
I needed this at work so I cleaned it up a bit. It was written in 2012 and was missing some documentation.
I added an example SQL CREATE statement for the database which makes it easier for users to get started with it.
I also removed the GetTables() in the Initialize because I don't think it's a good idea to do this every single request. It will crash anyway, and a database table won't just disappear...

I also moved the concatenation of the session_name with the session_id to a seperate function.

Let me know if this is ok.

### DIFF
--- a/includes/framework/QDbBackedSessionHandler.class.php
+++ b/includes/framework/QDbBackedSessionHandler.class.php
@@ -59,7 +59,7 @@
 		protected static $strSessionName	= ''; //PHPSESSID on a default PHP installation
 		
 		/** @var bool Whether to base64 the session data. Required when storing data in a TEXT field. */
-		public static $blnBase64			= false;
+		public static $blnBase64			= true;
 
 		/** @var bool Whether to compress the session data. */
 		public static $blnCompress			= true;
--- a/includes/framework/QDbBackedSessionHandler.class.php
+++ b/includes/framework/QDbBackedSessionHandler.class.php
@@ -11,7 +11,34 @@
 	 *  data - can be a BLOB or BINARY or VARBINARY or a TEXT. If TEXT, be sure to leave $blnBase64 on. If you are using
 	 * 		a binary field, you can turn that off to save space. Make sure your column is capable of holding the maximum size of
 	 * 		session data for your app, which depends on what you are putting in the $_SESSION variable.
+	 * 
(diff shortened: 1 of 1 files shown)


includes/framework/QDbBackedSessionHandler.class.php
```diff
@@ -48,17 +75,18 @@ class QDbBackedSessionHandler extends QBaseClass {
...
-			if (!array_key_exists(self::$intDbIndex, QApplication::$Database)) {
+			if (!isset(QApplication::$Database[self::$intDbIndex]) ) {
 				throw new QCallerException('No database defined at DB_CONNECTION index ' . self::$intDbIndex . '. Correct your settings in configuration.inc.php.');
 			}
-			$objDatabase = QApplication::$Database[self::$intDbIndex];
-			// see if the database contains a table with desired name
-			if (!in_array(self::$strTableName, $objDatabase->GetTables())) {
-				throw new QCallerException('Table ' . self::$strTableName . ' not found in database at DB_CONNECTION index ' . self::$intDbIndex . '. Correct your settings in configuration.inc.php.');
-			}
+// We don't want a GetTables everytime I do a request. This will slow the overal system down. Yes, it will crash in a slighty less clear way, does this matter?
```
spekary: Good

includes/framework/QDbBackedSessionHandler.class.php
```diff
@@ -29,13 +56,13 @@ class QDbBackedSessionHandler extends QBaseClass {
 		/**
 		 * @var string The session name to be used for saving sessions.
 		 */
-		protected static $strSessionName = '';
+		protected static $strSessionName	= ''; //PHPSESSID on a default PHP installation
 		
 		/** @var bool Whether to base64 the session data. Required when storing data in a TEXT field. */
-		public static $blnBase64 = true;
+		public static $blnBase64			= false;
```
spekary: I am concerned about this. This is potentially a breaking change, and somewhat difficult to debug. Can you leave it as true, since that will work for everyone and false is really just a step that optimizes on databases that can handle straight binary?

matthiaz:
I doubt this will break anything but I agree with the underlying thought. I changed it back to true. Sorry about the delay in the change.

kukrik:
Qualificatory question, where it merged the patch QCubed version? This is a good patch.

PR #1244, opened 2017-04-01
feature - Additional option to hide the the wrapper

kukrik:
Possible to get rid Tags (superfluous span or div) then use ShowWrapper = false; with only the function of the Render() or in order to avoid recurrent create custom work function (for example) RenderForBootrap($blnDisplayOutput = true) and ShowWrapper = false;  and so on.

It is designed for the custom theme etc...

Please see https://github.com/qcubed/qcubed/issues/1232

### DIFF
--- a/includes/base_controls/QBlockControl.class.php
+++ b/includes/base_controls/QBlockControl.class.php
@@ -42,6 +42,8 @@
 		protected $strTagName = null;
 		/** @var bool Should htmlentities be used on the contents of this control? */
 		protected $blnHtmlEntities = true;
+		/** @var bool Should the wrapper to turn off, for example, to use custom theme? */
+		protected $blnShowWrapper = true;
 
 		// BEHAVIOR
 		/** @var bool Is it a drop target? */
@@ -146,8 +148,9 @@
 		 * @return string The HTML string
 		 */
 		protected function GetControlHtml() {
+			if ($this->blnShowWrapper) {
 
-			$strToReturn = $this->RenderTag($this->strTagName,
+				$strToReturn = $this->RenderTag($this->strTagName,
 				null,
 				null,
 				$this->GetInnerHtml());
@@ -155,7 +158,12 @@
 //			if ($this->blnDropTarget)
 //				$strToReturn .= sprintf('<span id="%s_ctldzmask" style="position:absolute;"><span style="font-size: 1px">&nbsp;</span></span>', $this->strControlId);
 
-			return $strToReturn;
+				return $strToReturn;
+
+			} else {
+
+				return $this->GetInnerHtml();
+			}
 		}
 
 		/**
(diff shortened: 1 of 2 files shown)


kukrik:
I think that SwitchingWrapper is not quite accurate. Advanced idea is ShowWrapper. Can turn off or on, etc.

kukrik:
@spekary, @vaibhav-kaushal and boys! What do you think of this? Whether it is useful or not?

spekary:
Let me know what you are trying to accomplish. But this does not look correct. You are confused about what the wrapper is.

The wrapper is a span or div that surrounds any control. A block control is also a span or div. But they are treated differently. There is already a mechanism to turn off the wrapper. See the UseWrapper property.

A wrapper was part of the original QCubed functionality, and was the mechanism used to enable some ajax features, like drag and drop, and show and hide. Some of that functionality was replace by JQueryUI, and some of that by some changes in qcubed.js. However, there are specific situations where a wrapper is required to make things work:
1. When JQueryUI, or other javascript widgets, wrap a control with additional html. One situation is the JQuery UI dialog. The problem is that if we try to redraw the dialog, we only redraw the inner part and not the outer html that jQuery UI adds. The wrapper div makes sure the added html goes inside the wrapper, but outside the original control. They get sandwiched.
2. When a widget is drawn with a label, like RenderWithName. The wrapper makes sure both are contained, so when redrawing, both get drawn. And, if they get a drag and drop, they go together. The bootstrap plugin uses this.
3. When you need to show and hide a control that is a child of the form (rather than a child of another control). However, I fixed this recently in 3.1 to some extent. When a control is hidden, the wrapper is still drawn, but its invisible to the user. The wrapper is a placeholder so that if the control is shown again, we know where to put the control. My latest fix to 3.1 detects that there is no wrapper, and just adds the shown control to the end of the form, which works fine for some situations, like dialogs for instance.

A block control is specifically for outputting a div or span. It was not really designed for outputting just a group of html only.

I am not sure why you want to have something that would have no html wrapper at all. A div or span are very inexpensive, so not really sure the effort is important. Also, by having no tag at all, you kind of defeat the whole purpose of having QCubed. QCubed uses a control id to connect the block with javascript capabilities. Not having this will likely cause some bad things to happen in ajax drawing.

kukrik:
Ok, I understand, and I also agree with you. The question is that, earlier it was expensive development work and design theme of cheap work. It was understandable. In recent years, the growing importance of design share and cherish. It is inevitable and necessary.

At least good also important that clients are accepted the Bootstrap theme preferences that I've suggested. But some customers are price sensitive. And they bring to us purchased a custom theme. The client is king, and we have to accept :).

So avoid repetitive work, I have used in some places, for example, RenderWithName custom function.  I do not change much RenderWithName() standard function, the more it affects the CSS and layouts etc... The main thing is to work to quickly with clean templates.

I try to do differently. If the next time an interesting question, I ask again :). 

At present, I will lay this topic closed. Thank you for the debate!

PR #1262, opened 2017-05-17
InsertOrUpdate now returns the autoincremented ID

matthiaz:
InsertOrUpdate is a great method that is currently only used by the dbsessionhandler. It allows you to remove a query to determine if you want to insert or update because mysql will do this for you. 
Unfortunately, when you use this method, you didn't get the LAST_INSERT_ID on insert. I added this, plus when there is a duplicate, qcubed will return the id found in the update. 

To simplify the INSERT statement I also used the alternative INSERT syntax. INSERT INTO tablename SET col=val in stead of INSERT INTO tablename (col) values(val)

I'm running dbbackedsessionhandler in production.and it sees no difference between the old and new function. So there shouldn't be an impact on the rest of qcubed.

### DIFF
--- a/includes/database/QMySqliDatabase.class.php
+++ b/includes/database/QMySqliDatabase.class.php
@@ -52,20 +52,23 @@ if (!defined('MYSQLI_ON_UPDATE_NOW_FLAG')) {
 			return null;
 		}
 
-		public function InsertOrUpdate($strTable, $mixColumnsAndValuesArray, $strPKNames = null) {
-			$strEscapedArray = $this->EscapeIdentifiersAndValues($mixColumnsAndValuesArray);
+		public function InsertOrUpdate($strTable, $mixColumnsAndValuesArray, $strPKName = 'id') {
+			$strEscapedArray	= $this->EscapeIdentifiersAndValues($mixColumnsAndValuesArray);
 			$strUpdateStatement = '';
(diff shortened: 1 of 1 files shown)


spekary:
The (col) values(val) syntax was chosen because not all sql servers support the other syntax. That might have changed though. I don't know which are compliant and which are not.

Remember that currently all the sql we generate must work with all sql engines. For selects, its pretty standard across all of them. Updates and inserts are where a lot of sql engines diverge.

On the roadmap for v5 is changing things so that database drivers create their own sql based on the QQuery structure, which will allow whatever sql works for that driver to be used.

I suggest you do a quick look to see if all the modern sql engines support the new syntax, and then report back. Otherwise, you will need to change it to the old syntax for now.

matthiaz:
I didn't modify the generated INSERT statements. I considered it, but as you point out, the sql has to work for all sql engines. What I did change was the specific InsertOrUpdate method in QMySqliDatabase.class.php to make it a little more efficient. So that shouldn't have an impact on anything really. 

I did take up your advice and looked up the syntax for a couple of other databases (postgres, sql server,..). At first glance, it seems that only Mysql has support for this alternative syntax. So adding this to the generated code is a no go, I agree.

But I don't see the harm in changing the InsertOrUpdate method specifically for mysql.

PR #1284, opened 2017-10-26
feature - DisplayByteSize

kukrik:
Files or image file sizes in bytes are very uncomfortable to read or calculate. The function is created for human readability.
Example:
`$this->lblMessage->Text = 'Thanks for uploading the file: ' . $this->flaSample->FileName . ". File size: " . QString::DisplayByteSize($this->flaSample->Size, 2);`

### DIFF
--- a/includes/framework/QString.class.php
+++ b/includes/framework/QString.class.php
@@ -443,4 +443,26 @@
 				)
 				, 0, $intLength);
 		}
+
+		/**
+		 * Convert file size in bytes to human readable
+		 *
+		 * @param  integer  $intSize
+		 * @param  integer $intPrecision
+		 * @return  integer
+		 */
+		public static function DisplayByteSize($intSize, $intPrecision = null) {
+			if (is_null($intSize)) return 'n/a';
+			if ($intSize == 0) return '0 KB';
+			if ($intSize < 0) return '-';
+
+			if ($intSize > 0) {
+				$intSize = (int) $intSize;
+				$base = log($intSize) / log(1024);
+				$suffixes = array(' bytes', ' KB', ' MB', ' GB', ' TB');
+				return round(pow(1024, $base - floor($base)), $intPrecision) . $suffixes[floor($base)];
+			} else {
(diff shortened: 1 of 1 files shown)


kukrik:
At the moment, I was thinking that this feature could be further simplified. I'm giving one example:
```php
public static function FormatFileSize($intBytes) {
    if (is_null($intBytes)) return QApplication::Translate('n/a');
    if ($intBytes == 0) return QApplication::Translate('0 KB');
    if ($intBytes < 0) return QApplication::Translate('-');

    if ($intBytes >= 1073741824) {
        return round($intBytes / 1073741824, 2) . ' ' . QApplication::Translate('GB');
    }

    if ($intBytes >= 1048576) {
        return round($intBytes / 1048576, 2) . ' ' . QApplication::Translate('MB');
    }

    if ($intBytes >= 1024) {
        return round($intBytes / 1024, 2) . ' ' . QApplication::Translate('KB');
    }

    return $intBytes . ' ' . QApplication::Translate('bytes');
}
```
Or leave my first feature? What do you think?

matthiaz:
I like your second version better since it takes into account the translations. But I believe it is still too complex. 
How about something like this?

```
public static function ToHumanReadableFileSize($bytes, $decimals = 2) {
    $size = array('B','kB','MB','GB','TB','PB','EB','ZB','YB');
    $factor = floor((strlen($bytes) - 1) / 3);
    return sprintf("%.{$decimals}f", $bytes / pow(1024, $factor)) . QApplication::Translate($size[$factor]);
}
```

kukrik:
@matthiaz, your last-last solution seems to be elegant...

matthiaz:
@kukrik feel free to update your pull request to reflect this.

kukrik:
> @matthiaz, your last-last solution seems to be elegant...

The ToHumanReadableFileSize() function you are offering is much better. What do you think?

matthiaz:
@kukrik sorry about the late merge. But it's done

PR #1313, opened 2018-10-26
Ignore parent- and self-links when iterating Dir()

willypowwow:
Code fails attempting to unlink() the parent/self links.

### DIFF
--- a/includes/qform_state_handlers/QFileFormStateHandler.class.php
+++ b/includes/qform_state_handlers/QFileFormStateHandler.class.php
@@ -87,7 +87,7 @@
 			$objDirectory = dir(self::$StatePath);
 			while (($strFile = $objDirectory->read()) !== false) {
 				if (($strFile==".")||($strFile=="..")) continue;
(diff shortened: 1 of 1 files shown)


includes/qform_state_handlers/QFileFormStateHandler.class.php
```diff
@@ -86,7 +86,8 @@ public static function GarbageCollect() {
 			// Go through all the files
 			$objDirectory = dir(self::$StatePath);
 			while (($strFile = $objDirectory->read()) !== false) {
-				if (!count(self::$FileNamePrefix))
+				if (($strFile==".")||($strFile=="..")) continue;
+				if (!$FileNamePrefix)
```
matthiaz: @willypowwow I think you made and error there. This should be self:$FileNamePrefix right?

willypowwow: Yes, you are correct.  

```suggestion
				if (!$self::FileNamePrefix)
```

